PR #4, opened 2018-04-21
expose API to extract find_package-like CMake function calls

mikaelarguedas:
this allows for reuse in this package and for extension wanting to leverage that feature (e.g. https://github.com/eProsima/Fast-RTPS/pull/207#issuecomment-383248609)

### DIFF
--- a/colcon_cmake/package_identification/cmake.py
+++ b/colcon_cmake/package_identification/cmake.py
@@ -126,6 +126,41 @@ def _remove_cmake_comments_from_line(line):
     return modline
 
 
+def get_regex_extract_first_argument():
+    """
+    Return the regex used to extract the first argument of a function call.
+
+    The function call must be on a single line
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
spell checker was not happy with regex. It may be worth adding it to a whitelist ([this one](https://github.com/colcon/colcon.readthedocs.org/blob/master/spelling_wordlist.txt)?) if we think that's something we'll use "regular"ly

dirk-thomas:
After talking offline I pushed a modified patch which keeps the pattern, the regex evaluation and the extraction of the desired capture groups within a single function to avoid exposing too much information.

mikaelarguedas:
lgtm,
We could reuse that regex to find the project name. But as discussed offline we can do that separately.

Travis is unhappy though, looks like an install fluke

PR #5, opened 2018-04-25
add --cmake-target and --cmake-clean-first options

dirk-thomas:
Add new options:
* to build a specific target (`--cmake-build-target`) instead of the default ones and
* to `clean` before building (`--cmake-clean-first`).

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -40,10 +40,18 @@ class CmakeBuildTask(TaskExtensionPoint):
             '(args which start with a dash must be prefixed with an escaped '
             'space `\ `, e.g.: `--cmake-args \ -Dvar=val`)')
         parser.add_argument(
+            '--cmake-target',
+            help='Build a specific target instead of default targets.')
+        parser.add_argument(
             '--cmake-clean-cache',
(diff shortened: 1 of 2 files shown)


colcon_cmake/task/cmake/build.py
```diff
@@ -39,11 +39,19 @@ def add_arguments(self, *, parser):  # noqa: D102
...
+        parser.add_argument(
+            '--cmake-build-target',
+            help='Build a specific target instead of default targets.')
         parser.add_argument(
             '--cmake-clean-cache',
             action='store_true',
             help='Remove CMake cache before the build (implicitly forcing '
                  'CMake configure step)')
+        parser.add_argument(
+            '--cmake-clean-first',
```
mjcarroll: Nit: To be consistent with cmake's command flags, it may make sense to either do:
`--cmake-build-target` and `--cmake-build-clean-first`
OR
`--cmake-target` and `--cmake-clean-first`

dirk-thomas: The first options sound a bit too long. But the second one sounds good to me. Updated in 3be724656867bfe1adea3fdb26dc2d4eaed00c54.

Issue #6, opened 2018-04-25
fix help how to pass CMake args on Windows


sloretz:
Platform: windows 10
Shell: cmd
Installed via pip (freeze below)

The documentation for `--cmake-args` says arugments beginning with `-` must have an escaped space in front. It looks like escaping a space does not stop colcon's argument parser from trying to parse the cmake argument using the `cmd` shell on windows.

```
D:\overlay-ros2>colcon build --cmake-args \ -DCMAKE_BUILD_TYPE=Debug
usage: colcon [-h] [--log-level LOG_LEVEL]
              {build,extension-points,extensions,info,list,metadata,test,test-result}
              ...
colcon: error: unrecognized arguments: -DCMAKE_BUILD_TYPE=Debug
```

Quoting a space in front of the cmake argument works fine.
```
D:\overlay-ros2>colcon build --cmake-args " -DCMAKE_BUILD_TYPE=Debug"
```

Installed versions
```
C:\Users\osrf>pip freeze
attrs==17.4.0
colcon-cmake==0.1.1
colcon-common-extensions==0.1.1
colcon-core==0.1.0
colcon-defaults==0.1.0
colcon-devtools==0.1.0
colcon-library-path==0.1.0
colcon-metadata==0.1.0
colcon-notification==0.1.1
colcon-output==0.1.1
colcon-package-information==0.1.0
colcon-package-selection==0.1.0
colcon-parallel-executor==0.1.0
colcon-powershell==0.1.0
colcon-python-setup-py==0.1.0
colcon-recursive-crawl==0.1.0
colcon-ros==0.1.0
colcon-test-result==0.1.0
colorama==0.3.9
coloredlogs==9.0
coverage==4.5.1
cycler==0.10.0
empy==3.3.2
enum34==1.1.6
flake8==3.5.0
flake8-blind-except==0.1.1
flake8-builtins==1.0.post0
flake8-class-newline==1.6.0
flake8-comprehensions==1.4.1
flake8-deprecated==1.3
flake8-docstrings==1.3.0
flake8-import-order==0.17
flake8-polyfill==1.0.2
flake8-quotes==0.14.0
humanfriendly==4.10
matplotlib==2.0.2
mccabe==0.6.1
mock==2.0.0
nose==1.3.7
numpy==1.13.1
pbr==3.1.1
pep8==1.7.1
pluggy==0.6.0
py==1.5.2
pycodestyle==2.3.1
pydocstyle==2.1.1
[...]

PR #7, opened 2018-04-26
Document --cmake-args needs a quoted space before an argument with a dash on windows

sloretz:
The help text for `--cmake-args` now says to quote a space if using cmd on windows.

```
Arguments for 'cmake' packages:
  --cmake-args [* [* ...]]
                        Arbitrary arguments which are passed to all CMake
                        projects. Args that start with "-" must be prefixed
                        with a space. If using bash then use
                                --cmake-args \ -Dvar=val
                        If using Windows cmd then use
                                --cmake-args " -Dvar=val"
```

```
Arguments for 'cmake' packages:
  --ctest-args [* [* ...]]
                        Arbitrary arguments which are passed to all CTest
                        projects. Args that start with "-" must be prefixed
                        with a space. If using bash then use
                                --ctest-args \ -L label
                        If using Windows cmd then use
                                --ctest-args " -L" label
```

connects to colcon/colcon-cmake#6

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -36,10 +36,12 @@ class CmakeBuildTask(TaskExtensionPoint):
         parser.add_argument(
             '--cmake-args',
             nargs='*', metavar='*', type=str.lstrip,
-            help='Arbitrary arguments which are passed to all CMake projects '
-            '(Args that start with "-" must be prefixed with a space. '
-            'If using bash then use `\ `, e.g.: `--cmake-args \ -Dvar=val`. '
-            'If using Windows cmd then use: `--cmake-args " -Dvar=var"`)')
+            help='Arbitrary arguments which are passed to all CMake projects. '
+            'Args that start with "-" must be prefixed with a space. '
+            'If using bash then use'
+            '\n\t--cmake-args \ -Dvar=val\n'
+            'If using Windows cmd then use'
+            '\n\t--cmake-args " -Dvar=val"\n')
         parser.add_argument(
             '--cmake-clean-cache',
             action='store_true',
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -36,9 +36,9 @@ class CmakeBuildTask(TaskExtensionPoint):
         parser.add_argument(
             '--cmake-args',
             nargs='*', metavar='*', type=str.lstrip,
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
Does the same apply to `ctest-args` and other greedy arguments ?

sloretz:
@mikaelarguedas The same applies. I added documentation to `--ctest-args` in this PR and in colcon/colcon-ros#3 `--ament-cmake-args` and `--catkin-cmake-args`

mikaelarguedas:
Sounds good, I commented on another ticket with a suggestion to keep it simple and avoid calling out specific shells

sloretz:
@mikaelarguedas  ff9db414af20957c70e052cdba911caeea70547c

```
  --cmake-args [* [* ...]]
                        Arbitrary arguments which are passed to all CMake
                        projects. Args that start with "-" must be prefixed
                        with a space.
                                --cmake-args " -Dvar=val"
```

```
  --ctest-args [* [* ...]]
                        Arbitrary arguments which are passed to all CTest
                        projects. Args that start with "-" must be prefixed
                        with a space.
                                --ctest-args " -L" label
```

dirk-thomas:
The current help text is getting pretty long (4 lines). Let me try to suggest a shorter version which contains the improvements of this patch...

mikaelarguedas:
> The current help text is getting pretty long (4 lines)

It seems that this PR is decreasing the length of the help text rather than increasing it, though I agree that it would benefit from being even shorter :+1:

dirk-thomas:
How about the following which I picked for these reasons:
* Avoid unnecessary quotes around `"-"` - therefore using the word `dash`.
* Trying to clarify that each argument needs to be quoted separately.
* Keep indenting the example with not quotes / clutter around them.

```
  --cmake-args [* [* ...]]
                        Pass arguments to all CMake projects. Every arg
                        starting with a dash must be prefixed by a space, e.g.
                        	--cmake-args " -Dvar=val"
```

dirk-thomas:
With all the variations (CMake, CTest, Ament, Pytest) and their different lengths I came up with this wrapping instead which works consistently across all cases:

```
  --cmake-args [* [* ...]]
                        Pass arguments to all CMake projects. Every arg
                        starting with a dash must be prefixed by a space,
                        e.g. --cmake-args " -Dvar=val"
```

dirk-thomas:
After some offline iteration I committed a slightly more condensed version in 5cd23c6f562ea3767939bec6dc91e1034671b44e.

PR #12, opened 2018-05-15
use multiple processes by default with Visual Studio

dirk-thomas:
Fixes colcon/colcon-core#57.

Custom parameters can still be passed by setting the environment variable `CL` (similar to `MAKEFLAGS` for Make).

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -177,6 +177,10 @@ class CmakeBuildTask(TaskExtensionPoint):
             cmd += ['--clean-first']
         if is_multi_configuration_generator(args.build_base, args.cmake_args):
             cmd += ['--config', self._get_configuration(args)]
+
+            generator = get_generator(args.build_base)
+            if 'Visual Studio' in generator:
+                env = self._get_msbuild_environment(args, env)
         else:
             job_args = self._get_make_arguments()
             if job_args:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> What happens in the case where a user specify /m:2 ? Shouldn't we pass /MP2 (instead of just /MP) to respect the user specified value?

No, [/MP](https://msdn.microsoft.com/en-us/library/bb385193.aspx) only indicates to build with multiple processes. [/m:X](https://msdn.microsoft.com/en-us/library/ms164311.aspx) where `X` specifies the maximum number of concurrent processes.

mikaelarguedas:
> No, /MP only indicates to build with multiple processes

From the /MP page you linked:
"If you omit the processMax argument, the compiler retrieves the number of effective processors on your computer from the operating system, and creates a process for each processor."
"The number argument of the /maxcpucount MSBuild option specifies the maximum number of projects to build at the same time. Similarly, the processMax argument of the /MP compiler option specifies the maximum number of compilation units to build at the same time. If the /maxcpucount option specifies P projects and the /MP option specifies C processes, a maximum of PxC processes execute at the same time."

dirk-thomas:
I modified the condition to only check for `/MP` in https://github.com/colcon/colcon-cmake/pull/12/files#diff-af9d22be861287184d5eedb3f47a01daR212

dirk-thomas:
One example build on Windows using all available cores: [Build Status](https://ci.ros2.org/job/ci_windows/4473/)

PR #19, opened 2018-06-19
fix CMAKE_MODULE_PATH env var

dirk-thomas:
Currently the environment hooks are being generated in (depending in where the CMake module was found):

```
<prefix>/share/<pkgname>/cmake/Modules/share/<pkgname>/hook/cmake_module_path.bat
```

and the appended path points to:

```
<prefix>
```

With this patch the location of the environment hook is not nested anymore:

```
<prefix>/share/<pkgname>/hook/cmake_module_path.bat
```

and the appended path is the directory containing the CMake module:

```
<prefix>/share/<pkgname>/cmake/Modules
```

The first fix is just cosmetic. The second one avoid cross-talk between packages (one extending the module path and another one benefiting from it).

Additionally the patch fixes the case where a package installs multiple CMake module into different location. Before the last hook was overwriting all previous ones.

### DIFF
--- a/colcon_cmake/environment/cmake_module_path.py
+++ b/colcon_cmake/environment/cmake_module_path.py
@@ -10,7 +10,7 @@ from colcon_core.shell import create_environment_hook
 
 
 class CmakeModulePathEnvironment(EnvironmentExtensionPoint):
-    """Extend the `CMAKE_PREFIX_PATH` variable to find CMake modules."""
+    """Extend the `CMAKE_MODULE_PATH` variable to find CMake modules."""
 
     def __init__(self):  # noqa: D107
         super().__init__()
@@ -21,12 +21,17 @@ class CmakeModulePathEnvironment(EnvironmentExtensionPoint):
         hooks = []
 
         logger.log(1, "checking '%s' for CMake module files" % prefix_path)
+        count = 0
         for dirpath, _, filenames in os.walk(str(prefix_path)):
             for filename in filenames:
                 if filename.startswith('Find') and filename.endswith('.cmake'):
+                    count += 1
                     hooks += create_environment_hook(
-                        'cmake_module_path', prefix_path / dirpath,
-                        pkg_name, 'CMAKE_MODULE_PATH', '', mode='prepend')
+                        'cmake_module_path' +
+                        (str(count) if count > 1 else ''),
+                        prefix_path, pkg_name, 'CMAKE_MODULE_PATH',
+                        os.path.relpath(dirpath, start=str(prefix_path)),
+                        mode='prepend')
                     break
 
         return hooks


dirk-thomas:
In a non-isolated install this bug also results in numerous extra environment hooks to be created.

colcon_cmake/environment/cmake_module_path.py
```diff
@@ -21,12 +21,17 @@ def create_environment_hooks(self, prefix_path, pkg_name):  # noqa: D102
...
+        count = 0
         for dirpath, _, filenames in os.walk(str(prefix_path)):
             for filename in filenames:
                 if filename.startswith('Find') and filename.endswith('.cmake'):
+                    count += 1
                     hooks += create_environment_hook(
-                        'cmake_module_path', prefix_path / dirpath,
-                        pkg_name, 'CMAKE_MODULE_PATH', '', mode='prepend')
+                        'cmake_module_path' +
+                        (str(count) if count > 1 else ''),
```
wjwwood: I don't fully understand why this needs to have a number appended every time, won't you generate a lot of redundant environment hooks if there are several find modules in the same location?

dirk-thomas: The outer loop iterates over different directories and the inner loop iterates over the files in one directory. For each directory only one hook can be generated (hence a [break](https://github.com/colcon/colcon-cmake/pull/19/files#diff-90c1e311baa57d65b1dfab53555fef91R35) after the generation in the inner loop). For a different directory another hook can be generated though which would point to a different directory.

To make an example for a package `foo` providing CMake modules in two different locations there would now be two hooks which prepend two different directories to `CMAKE_MODULE_PATH`:

* hook `cmake_module_path.sh` would prepend `<prefix>/share/foo/some/location` (even if there are multiple CMake modules in this directory only one hook is generated)
* hook `cmake_module_path2.sh` would prepend `<prefix>/share/foo/other/location`

Does that make sense?

wjwwood: Yes 👍

dirk-thomas:
So the idea was good and it works well in isolated installations. But...

In the `--merge-install` case this is still a problem. After a package A installed its CMake module (and got the intended hook for it) a later processed package B will find the same CMake module and also generate a hook for it. While the presence of the extra hook doesn't do any harm to the environment (since the entry is already there it will be skipped) it does add a lot of unnecessary hooks which takes extra time for sourcing the setup files.

I think the only "reliable" option is to limit the search to location which are considered by [CMake](https://cmake.org/cmake/help/latest/command/find_package.html) as well as contain the package name in the path. I will create a follow up PR for this after confirming this works for our cases...

PR #21, opened 2018-08-24
Add recommends to deb package

Theosakamg:
Add to debian package recommends for python3-colcon-argcomplete

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,4 +1,5 @@
 [colcon-cmake]
 Depends3: python3-colcon-core, python3-colcon-library-path
+Recommends3: python3-colcon-argcomplete
 Suite: xenial bionic stretch buster
 X-Python3-Version: >= 3.5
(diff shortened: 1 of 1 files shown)


stdeb.cfg
```diff
@@ -1,4 +1,5 @@
 [colcon-cmake]
 Depends3: python3-colcon-core, python3-colcon-library-path
+Recommends3: python3-colcon-argcomplete, cmake
```
dirk-thomas: I don't think `python3-colcon-argcomplete` should be recommended for this package. This package might be used in an environment which doesn't need completion (e.g. CI) or the user might not use a shell supported by `colcon-argcomplete`.

In my mind `colcon-common-extensions` was the package recommended to install if you wants a collections of "useful common" extensions. Otherwise the user should be able to install only the Debian packages needed (without anything being pulled in which they don't want).

PR #22, opened 2018-09-12
Update to use DependencyDescriptor

murphm8:
This is an update for an upcoming change to colcon-core: colcon/colcon-core#107

Related to: colcon/colcon-core#106


murphm8:
This doesn't add any additional metadata beyond the name of the dependency so I'm going to close this for now since DependencyDescriptor is backwards compatible with str.

PR #24, opened 2018-10-09
Match against capital PROJECT command

mxgrey:
CMake allows commands to be all lowercase *or* all uppercase, so we should check for a capital `PROJECT` in addition to lowercase `project`, or else we may miss some valid cases.

### DIFF
--- a/colcon_cmake/package_identification/cmake.py
+++ b/colcon_cmake/package_identification/cmake.py
@@ -140,7 +140,7 @@ def extract_project_name(content):
     # extract project name
     match = re.search(
         # keyword
-        'project'
+        'project|PROJECT'
         # optional white space
         r'\s*'
         # open parenthesis
--- a/colcon_cmake/package_identification/cmake.py
+++ b/colcon_cmake/package_identification/cmake.py
@@ -140,7 +140,7 @@ def extract_project_name(content):
     # extract project name
     match = re.search(
         # keyword
-        '[p|P][r|R][o|O][j|J][e|E][c|C][t|T]'
+        'project'
         # optional white space
         r'\s*'
         # open parenthesis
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> or else we may miss some valid cases.

Since CMake even allows mixed case for function names we might want to allow that too.

dirk-thomas:
The same applies to the other regulsr expressions, e.g. for finding dependencies.

mxgrey:
I'm not very experienced with regex, so the latest commit kind of brute-forces an attempt at case insensitivity. If anyone knows a more elegant approach, please do share.

mxgrey:
It seems Python `re` offers an argument flag `re.IGNORECASE`, so I've switched the searches to use that.

The only potential pitfall I can think of is that the search for `pkg_check_modules` also looks for a match against `REQUIRED|QUIET|NO_CMAKE_PATH|NO_CMAKE_ENVIRONMENT_PATH`, but now the case for those will be ignore, even though the arguments are technically case-sensitive. I don't imagine that will cause anyone any practical issues, though.

dirk-thomas:
> It seems Python re offers an argument flag `re.IGNORECASE`, so I've switched the searches to use that.
> The only potential pitfall I can think of is that the search for `pkg_check_modules` also looks for a match against `REQUIRED|QUIET|NO_CMAKE_PATH|NO_CMAKE_ENVIRONMENT_PATH`, ...


Making the whole regex case insensitive isn't an option for the reason you already mentioned.

Unfortunately making parts of the regex case insensitive with something like this (`(?i:foo)bar`) is only available as of Python 3.6 :disappointed:

dirk-thomas:
I created #25 which uses the regex pattern available as of Python 3.6 when possible and falls back to expanding the string into lowercase / uppercase options. It also applies the patch to all matched function names.

PR #25, opened 2018-10-09
match case insensitive function names

dirk-thomas:
Replaces #24.


dirk-thomas:
> I might suggest renaming `_get_pattern(~)` to `_get_case_insensitive_pattern(~)`

:+1: b7513abf9dace926f19937066f4efe24e0b5af4d

PR #31, opened 2019-02-01
allow additional targets for CMake build

dirk-thomas:
Enabling colcon/colcon-ros#48.

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -64,7 +64,7 @@ class CmakeBuildTask(TaskExtensionPoint):
 
     async def build(
         self, *, additional_hooks=None, skip_hook_creation=False,
-        environment_callback=None
+        environment_callback=None, additional_targets=None
     ):  # noqa: D102
         pkg = self.context.pkg
         args = self.context.args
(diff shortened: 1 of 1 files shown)


colcon_cmake/task/cmake/build.py
```diff
@@ -189,32 +190,49 @@ def _store_cmake_args(self, build_base, cmake_args):
...
-        else:
-            job_args = self._get_make_arguments()
-            if job_args:
-                cmd += ['--'] + job_args
-        return await check_call(
-            self.context, cmd, cwd=args.build_base, env=env)
+
+        for i, target in enumerate(targets):
+            if args.cmake_target_skip_unavailable:
+                if not await has_target(args.build_base, target):
```
cottsay: How does this work if `target` is an empty string, which I'm assuming means "default targets", and `cmake_target_skip_unavailable` is `False`? I don't see anything in `has_target` for handling an empty string.

dirk-thomas: Nice catch, that certainly does not work. Should be fixed in 17844894074d798144a54a25ebe9eb4f2fdc038f.

colcon_cmake/task/cmake/build.py
```diff
@@ -189,32 +190,48 @@ def _store_cmake_args(self, build_base, cmake_args):
...
+                cmd += ['--clean-first']
+            if multi_configuration_generator:
+                cmd += ['--config', self._get_configuration(args)]
+            else:
+                job_args = self._get_make_arguments()
+                if job_args:
+                    cmd += ['--'] + job_args
+            rc = await check_call(
+                self.context, cmd, cwd=args.build_base, env=env)
+            if rc:
```
dirk-thomas: This check is incorrect: see #32.

PR #32, opened 2019-02-08
fix check of return code

dirk-thomas:
Follow up of #31.


paulbovbel:
LGTM, thanks!

colcon_cmake/task/cmake/build.py
```diff
@@ -230,8 +230,8 @@ def _get_last_cmake_args_path(self, build_base):
                     cmd += ['--'] + job_args
             rc = await check_call(
                 self.context, cmd, cwd=args.build_base, env=env)
-            if rc:
-                return rc
+            if rc and rc.returncode:
```
dirk-thomas: While the condition needs to check `rc.returncode` the internal function should return just `rc` since the calling code is expecting a `subprocess.CompletedProcess`. See #33.

PR #33, opened 2019-02-11
fix return type

dirk-thomas:
Follow up of #32.

@paulbovbel FYI

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -231,7 +231,7 @@ class CmakeBuildTask(TaskExtensionPoint):
             rc = await check_call(
                 self.context, cmd, cwd=args.build_base, env=env)
             if rc and rc.returncode:
-                return rc.returncode
+                return rc
 
     def _get_configuration(self, args):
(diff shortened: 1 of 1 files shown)


paulbovbel:
Isn't this basically what was there before https://github.com/colcon/colcon-cmake/commit/7ac46cc59032682babbb95ad2ba30c1f89b755fe?

dirk-thomas:
No.

Initially it was checking for `if rc` which was always `True` since `rc` is an instance of `subprocess.CompletedProcess`. So the loop always stopped after the first target.

With the recent change I fixed the condition to check `if rc and rc.returncode` but by then returning `rc.returncode` the caller of the function wasn't happy which still expects that if something is returned it to be a `subprocess.CompletedProcess`.

paulbovbel:
Thanks, I'll update our system once you cut a release.

dirk-thomas:
> I'll update our system once you cut a release.

:ship:

MatteoRagni:
May I suggest to update the title of this pull request with something a little more easy to find? Something like **Fix Return type (rc is of type int in build.py)**

PR #41, opened 2019-05-17
add test result extension for CTest result files

### LINKED ISSUE #40: support reading CTest result files
Support reading `build/<pkgname>/Testing/<timestamp>/Test.xml` files and showing them in the `test-result` verb. This will require the `test-result` verb to provide an extension point for external result parsers.

dirk-thomas:
Closes #40. Requires colcon/colcon-test-result#20.

Before merging a version dependency on the upcoming `colcon-test-result` package needs to be added.

### DIFF
--- a/colcon_cmake/test_result/ctest.py
+++ b/colcon_cmake/test_result/ctest.py
@@ -28,13 +28,9 @@ class CtestTestResult(TestResultExtensionPoint):
 
     def get_test_results(self, basepath, *, collect_details):  # noqa: D102
         results = set()
-        # only consider direct subdirectories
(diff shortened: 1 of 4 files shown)


colcon_cmake/test_result/ctest.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2019 Dirk Thomas
+# Licensed under the Apache License, Version 2.0
+
+from xml.etree.ElementTree import ElementTree
```
mjcarroll: Thanks for not using `lxml` ;)

PR #43, opened 2019-05-20
fix variable name in warning

dirk-thomas:
Fixes exception reported in colcon/colcon-core#189.


dirk-thomas:
@mangoSteve Please comment here if the patch resolves the exception in your case.

mangoSteve:
Thank you so much, this DO SOLVE the problem.

PR #46, opened 2019-06-03
support CMake -A option to select architecture on Windows

### LINKED ISSUE #45: Option to select target platform (e.g. "Visual Studio 15 2017" 32bit)
Hi,

I'm coming here from https://answers.ros.org/question/324217/ros2-for-win32-x86-how-to-make-colcon-call-cmake-without-win64/

I need to build ROS2 for Win32 and I am missing the option to select the cmake target platform "Visual Studio 15 2017". Is it possible to pass this as an argument? Or could I just remove the "Win64" here: https://github.com/colcon/colcon-cmake/blob/14ccfdc4aab94e2376d46c3906d7fc5d0c797c4b/colcon_cmake/task/cmake/build.py#L147-L162 ?

Thanks
marosso

dirk-thomas:
Closes #45.

For VS 15 and older the default architecture is `Win32`. Instead of hard coding `Win64` as part of the generator name this patch passes the `-A` option and also allows the user to pass a custom value.

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -152,14 +152,19 @@ class CmakeBuildTask(TaskExtensionPoint):
                     'please run within a Visual Studio Command Prompt.')
             supported_vsv = {
                 '16.0': 'Visual Studio 16 2019',
-                '15.0': 'Visual Studio 15 2017 Win64',
-                '14.0': 'Visual Studio 14 2015 Win64',
+                '15.0': 'Visual Studio 15 2017',
+                '14.0': 'Visual Studio 14 2015',
             }
             if vsv not in supported_vsv:
                 raise RuntimeError(
                     "Unknown / unsupported VS version '{vsv}'"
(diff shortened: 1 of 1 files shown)


colcon_cmake/task/cmake/build.py
```diff
@@ -152,14 +152,19 @@ def add_arguments(self, *, parser):  # noqa: D102
...
             }
             if vsv not in supported_vsv:
                 raise RuntimeError(
                     "Unknown / unsupported VS version '{vsv}'"
                     .format_map(locals()))
             cmake_args += ['-G', supported_vsv[vsv]]
+            # choose 'x64' on VS 14 and 15 if not specified explicitly
+            # since otherwise 'Win32' is the default for those
+            # newer versions default to the host architecture
+            if '-A' not in args.cmake_args and vsv in ('14.0', '15.0'):
```
dirk-thomas: This line raises if `args.cmake_args` is `None`.

PR #53, opened 2019-10-15
Provide options for generator override on Windows.

seanyen:
Beginning from [`Visual Studio 2017 15.4`](https://devblogs.microsoft.com/cppblog/cmake-support-in-visual-studio/), the toolchain comes with `CMake` and `Ninja` built-in support. It would be great to see Windows developers can have the free choice on what generator to use. This pull request is to unblock the hard-coded decision in `colcon-cmake` and allow developers to pass `--cmake-args -G Ninja` to specify the generator by their choices.


dirk-thomas:
Thanks for the patch.

PR #56, opened 2019-10-20
cmake_install always invokes --build --target install

KazNX:
Removed special case code for Visual Studio generator using MSBUILD. "cmake --build" with "--target install" has the same effect and removed the need for special case code.

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -321,14 +321,8 @@ class CmakeBuildTask(TaskExtensionPoint):
         if multi_configuration_generator:
             cmd += ['--config', self._get_configuration(args)]
         else:
-            # Add make/ninja arguments for the number of job threads to use.
             job_args = self._get_make_arguments()
             if job_args:
-                # Add arguments for the number of job threads to use
-                # Arguably, _get_make_arguments() should be be configured to
-                # return arguments based on the current generator, not just for
-                # make (and ninja by coincidence). The '--' extension may be
-                # better returned from _get_make_arguments()
                 cmd += ['--'] + job_args
         return await check_call(
             self.context, cmd, cwd=args.build_base, env=env)
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -313,27 +313,11 @@ class CmakeBuildTask(TaskExtensionPoint):
     async def _install(self, args, env):
         self.progress('install')
 
-        generator = get_generator(args.build_base)
-        if 'Visual Studio' not in generator:
-            if CMAKE_EXECUTABLE is None:
-                raise RuntimeError("Could not find 'cmake' executable")
-            cmd = [
+        if CMAKE_EXECUTABLE is None:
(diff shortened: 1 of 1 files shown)


colcon_cmake/task/cmake/build.py
```diff
@@ -313,27 +311,15 @@ def _get_make_arguments(self):
...
         self.progress('install')
 
-        generator = get_generator(args.build_base)
-        if 'Visual Studio' not in generator:
-            if CMAKE_EXECUTABLE is None:
-                raise RuntimeError("Could not find 'cmake' executable")
-            cmd = [
-                CMAKE_EXECUTABLE, '--build', args.build_base,
-                '--target', 'install']
-            job_args = self._get_make_arguments()
```
dirk-thomas: This line has been removed which needs to be restored.

colcon_cmake/task/cmake/build.py
```diff
@@ -313,27 +311,15 @@ def _get_make_arguments(self):
...
-            if job_args:
-                cmd += ['--'] + job_args
```
dirk-thomas: These three lines have been removed which need to be restored.

KazNX: `job_args` restored when not using a multi-configuration generator. Seems the usage of job args is similarly dependent on the multi-configuration generator check in the build step. The installation code as is may not have worked for Xcode.

colcon_cmake/task/cmake/build.py
```diff
@@ -313,27 +311,24 @@ def _get_make_arguments(self):
...
+        multi_configuration_generator = is_multi_configuration_generator(
+            args.build_base, args.cmake_args)
+        if multi_configuration_generator:
+            cmd += ['--config', self._get_configuration(args)]
+        else:
+            # Add make/ninja arguments for the number of job threads to use.
             job_args = self._get_make_arguments()
             if job_args:
+                # Add arguments for the number of job threads to use
+                # Arguably, _get_make_arguments() should be be configured to
```
dirk-thomas: If you think this should be done please consider to contribute a pull request for it. This comment in the code is unlikely to be revisited hence I will remove it.

colcon_cmake/task/cmake/build.py
```diff
@@ -313,27 +311,24 @@ def _get_make_arguments(self):
...
+                # return arguments based on the current generator, not just for
+                # make (and ninja by coincidence). The '--' extension may be
+                # better returned from _get_make_arguments()
```
dirk-thomas: The `--` has nothing to do with the make arguments but with the fact that the invoked cmd is `cmake --build` and arguments which should be passed along to the specific tool need to be separated by a double dash. Therefore I will remove this comment.

dirk-thomas:
CI builds using this branch:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8515/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4398/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6932/) (known failing test)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8428/)

PR #57, opened 2019-10-21
Use 'cmake --install' where available

KazNX:
Make use of CMake '--install' option introduced in 3.15. Only performs installation, whereas '--build <dir> --target install' would first (re)build. Using '--install' is a significantly faster installation process when using Visual Studio as various overheads are avoided.

### DIFF
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -229,25 +229,27 @@ def _parse_cmake_version():
     :returns: The version as parsed by the pkg_resources package
     :rtype pkg_resources.extern.packaging.version.Version
     """
-    output = subprocess.run([CMAKE_EXECUTABLE, '--version'],
-                            capture_output=True)
-    if output.returncode:
-        # Failed to execute 'cmake --version'
+    try:
+        output = subprocess.check_output([CMAKE_EXECUTABLE, '--version'])
+        lines = output.decode().splitlines()
+        ver_line = lines[0] if lines and len(lines) else None
+        if ver_line:
+            # Extract just the version part of the string.
+            ver_re_str = r'^(?:.*)(\d+\.\d+\..*)'
+            ver_match = re.match(ver_re_str, ver_line, re.I)
+            if ver_match:
+                return parse_version(ver_match.group(1))
         return None
-    lines = output.stdout.decode().splitlines()
-    ver_line = lines[0] if lines and len(lines) else None
-    if ver_line:
-        # Extract just the version part of the string.
-        ver_re_str = r'^(?:.*)(\d+\.\d+\..*)'
-        ver_match = re.match(ver_re_str, ver_line, re.I)
-        if ver_match:
-            return parse_version(ver_match.group(1))
-    # Failed to parse version number.
+    except subprocess.CalledProcessError:
+        return None
     return None
 
 
 # Global variable for the cached CMake version number.
-_cached_cmake_version = _parse_cmake_version()
+# When valid, this will be of pkg_resources.extern.packaging.version.Version
+# It may also have a boolean value False when the cmake version has failed
+# to parse. This saves recurring parse failures.
+_cached_cmake_version = None
 
 
 def get_cmake_version():
@@ -261,4 +263,19 @@ def get_cmake_version():
     :returns: The version as parsed by the pkg_resources package
     :rtype pkg_resources.extern.packaging.version.Version
     """
-    return _cached_cmake_version
+    global _cached_cmake_version
+    if _cached_cmake_version is None:
+        # No value yet. Parse the version number.
+        _cached_cmake_version = _parse_cmake_version()
+        if not _cached_cmake_version is None:
+            # Success.
+            return _cached_cmake_version
+        # Failed to parse. Set _cached_cmake_version to False to prevent
+        # parsing again, but we can still return None as required.
+        _cached_cmake_version = False
+    # Have cached a previous result. Check if the previous result is a bool
+    # type. If so, this implies we've tried and failed to parse the version
+    # number and should return None. Otherwise return the cached value as is.
+    elif not isinstance(_cached_cmake_version, bool):
+        return _cached_cmake_version
+    return None
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -221,10 +221,10 @@ def _parse_cmake_version(version_str):
     :rtype (int, int, int, str)
     """
     # Version number is on the first line. We can ignore anything else.
+    ver_re_str = \
+        r'^(?:cmake\s+version)?\s+([0-9]+)\.([0-9]+)\.([0-9]+)(?:\-(.*))?\s*.*$'
     if version_str and len(version_str):
-        ver_match = re.match(
-            r'^(?:cmake\s+version)?\s+([0-9]+)\.([0-9]+)\.([0-9]+)(?:\-(.*))?\s*.*$',
-            version_str)
+        ver_match = re.match(ver_re_str, version_str)
         if ver_match:
             return (int(ver_match.group(1)), int(ver_match.group(2)),
                     int(ver_match.group(3)), ver_match.group(4))
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -240,26 +240,6 @@ def get_cmake_version():
     return _cached_cmake_version
 
 
-def parse_cmake_version_string(version_string):
-    """
-    Parse the given CMake version string.
-
-    This is for internal use and testing.
-
-    Expects strings of the form 'cmake version 3.15.1'.
-
-    :param str version_string: The version string to parse.
-    :returns: The parsed version string or None on failure to parse.
-    :rtype pkg_resources.extern.packaging.version.Version
-    """
-    # Extract just the version part of the string.
-    ver_re_str = r'^(?:.*[ \t])?(\d+\.\d+\.\d+).*'
-    ver_match = re.match(ver_re_str, version_string)
-    if ver_match:
-        return parse_version(ver_match.group(1))
-    return None
-
-
 def _parse_cmake_version():
     """
     Parse the CMake version printed by `CMAKE_EXECUTABLE --version`.
@@ -277,5 +257,23 @@ def _parse_cmake_version():
         lines = output.decode().splitlines()
         if lines:
             # Parse just the version part of the string.
-            return parse_cmake_version_string(lines[0])
+            return _parse_cmake_version_string(lines[0])
+    return None
+
+
+def _parse_cmake_version_string(version_string):
+    """
+    Parse the given CMake version string.
+
+    Expects strings of the form 'cmake version 3.15.1'.
+
+    :param str version_string: The version string to parse.
+    :returns: The parsed version string or None on failure to parse.
+    :rtype pkg_resources.extern.packaging.version.Version
+    """
+    # Extract just the version part of the string.
+    ver_re_str = r'^(?:.*[ \t])?(\d+\.\d+\.\d+).*'
+    ver_match = re.match(ver_re_str, version_string)
+    if ver_match:
+        return parse_version(ver_match.group(1))
     return None
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -215,24 +215,22 @@ def get_visual_studio_version():
 """
 Global variable for the cached CMake version number.
 
-When valid, this will be of pkg_resources.extern.packaging.version.Version
-It may also have a boolean value False when the CMake version has failed
-to parse.
+When valid, this will be a pkg_resources.extern.packaging.version.Version.
+It may also have a boolean value False when the CMake version could not be
+determined to avoid trying to determine it again.
 """
 _cached_cmake_version = None
 
 
 def get_cmake_version():
     """
-    Get the cached CMake version number if available and successfully parsed.
+    Get the CMake version.
 
-    The result is None when a version number is not available or cannot be
-    parsed.
+    The function caches the result on the first invocation and reused that on
+    subsequent invocations.
+    The result is None when a version number could not be determined.
 
-    This may occur if CMake is not present, is a very old version or if a newer
-    changes the output format.
-
-    :returns: The version as parsed by the pkg_resources package
+    :returns: The version as reported by `CMAKE_EXECUTABLE --version`, or None
     :rtype pkg_resources.extern.packaging.version.Version
     """
     global _cached_cmake_version
@@ -262,23 +260,9 @@ def get_cmake_version():
 
 def _parse_cmake_version():
     """
-    Parse the CMake version number by running `cmake --version`.
+    Parse the CMake version printed by `CMAKE_EXECUTABLE --version`.
 
-    The version number is parse from the output of 'cmake --version' and is
-    expected in the form 'cmake version #.#.#' where each '#' character
-    represents part of the version number.
-
-    Additional text may follow, however only the first line is parsed.
-
-    Parsing is performed by the pkg_resources package and the returned object
-    is a Version object from that package.
-
-    This function blocks on the execution of cmake and should only be used to
-    cache the CMake version number.
-
-    External API users should use get_cmake_version()
-
-    :returns: The version as parsed by the pkg_resources package
+    :returns: The version parsed by pkg_resources.parse_version, or None
     :rtype pkg_resources.extern.packaging.version.Version
     """
     try:
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -263,7 +263,7 @@ def parse_cmake_version_string(version_string):
     :rtype pkg_resources.extern.packaging.version.Version
     """
     # Extract just the version part of the string.
-    ver_re_str = r'^.*(\d+\.\d+\.\d+).*'
+    ver_re_str = r'^(?:.*[ \t])?(\d+\.\d+\.\d+).*'
     ver_match = re.match(ver_re_str, version_string)
     if ver_match:
         return parse_version(ver_match.group(1))
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -212,9 +212,9 @@ def get_visual_studio_version():
     return os.environ.get('VisualStudioVersion', None)
 
 
-async def get_cmake_version():
+def _parse_cmake_version():
     """
-    Get the version number of CMake.
+    Parse the CMake version number by running `cmake --version`.
 
     The version number is parse from the output of 'cmake --version' and is
     expected in the form 'cmake version #.#.#' where each '#' character
@@ -222,11 +222,19 @@ async def get_cmake_version():
     only the first line is parsed. Parsing is performed by the pkg_resources
     package and the returned object is a Version object from that package.
 
+    This function blocks on the execution of cmake and should only be used to
+    cache the CMake version number. External API users should use
+    get_cmake_version()
+
     :returns: The version as parsed by the pkg_resources package
     :rtype pkg_resources.extern.packaging.version.Version
     """
-    output = await check_output([CMAKE_EXECUTABLE, '--version'])
-    lines = output.decode().splitlines()
+    output = subprocess.run([CMAKE_EXECUTABLE, '--version'],
+                            capture_output=True)
+    if output.returncode:
+        # Failed to execute 'cmake --version'
+        return None
+    lines = output.stdout.decode().splitlines()
     ver_line = lines[0] if lines and len(lines) else None
     if ver_line:
         # Extract just the version part of the string.
@@ -236,3 +244,15 @@ async def get_cmake_version():
             return parse_version(ver_match.group(1))
     # Failed to parse version number.
     return None
+
+# Global variable for the cached CMake version number.
+_cached_cmake_version = _parse_cmake_version()
+def get_cmake_version():
+    """
+    Get the cached CMake version number or None if the version number is not
+    available and parsable.
+
+    :returns: The version as parsed by the pkg_resources package
+    :rtype pkg_resources.extern.packaging.version.Version
+    """
+    return _cached_cmake_version
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -234,5 +234,5 @@ async def get_cmake_version():
         ver_match = re.match(ver_re_str, ver_line, re.I)
         if ver_match:
             return parse_version(ver_match.group(1))
-    # Failed to extract parsable version number.
+    # Failed to parse version number.
     return None
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -236,27 +236,18 @@ def get_cmake_version():
     """
     global _cached_cmake_version
     if _cached_cmake_version is None:
-        # No value yet. Parse the version number.
-        _cached_cmake_version = _parse_cmake_version()
-        if _cached_cmake_version is not None:
-            # Success.
-            return _cached_cmake_version
-        # Failed to parse.
-        #
-        # Set _cached_cmake_version to False to prevent parsing again, but we
-        # can still return None as required.
-        _cached_cmake_version = False
-    # Have cached a previous result.
-    #
-    # Check if the previous result is a bool type.
-    #
-    # If so, this implies we've tried and failed to parse the version
-    # number and should return None.
-    #
-    # Otherwise return the cached value as is.
-    elif not isinstance(_cached_cmake_version, bool):
-        return _cached_cmake_version
-    return None
+        cmake_version = _parse_cmake_version()
+        if cmake_version is not None:
+            _cached_cmake_version = cmake_version
+        else:
+            # flag to avoid retrying on subsequent calls
+            _cached_cmake_version = False
+
+    if _cached_cmake_version is False:
+        # in case the CMake version could not be determined
+        return None
+
+    return _cached_cmake_version
 
 
 def _parse_cmake_version():
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -2,13 +2,13 @@
 # Licensed under the Apache License, Version 2.0
 
 import os
-from pkg_resources import parse_version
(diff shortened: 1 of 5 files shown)


dirk-thomas:
Builds on top of #56. Waiting for that one to be merged first.

dirk-thomas:
With #56 merged can you please rebase this patch.

KazNX:
I've resolve the merge conflicts, however, I merged rather than rebasing.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +209,53 @@ def get_visual_studio_version():
...
+    ver_re_str = \
+        r'^(?:cmake\s+version)?\s+([0-9]+)\.([0-9]+)\.([0-9]+)(?:\-(\S+))?'
+    if version_str and len(version_str):
+        ver_match = re.match(ver_re_str, version_str)
+        if ver_match:
+            return (int(ver_match.group(1)), int(ver_match.group(2)),
+                    int(ver_match.group(3)), ver_match.group(4))
+
+    # Failed to parse the version number.
+    return (0, 0, 0, 'failed')
```
dirk-thomas: Instead of implementing a custom function how about using `pkg_resources.parse_version` instead?

colcon_cmake/task/cmake/build.py
```diff
@@ -313,16 +314,30 @@ def _get_make_arguments(self):
 
         if CMAKE_EXECUTABLE is None:
             raise RuntimeError("Could not find 'cmake' executable")
-        cmd = [
-            CMAKE_EXECUTABLE, '--build', args.build_base,
-            '--target', 'install']
+        cmake_ver = await get_cmake_version()
```
dirk-thomas: Currently the return value is not being checked for a failure.

colcon_cmake/task/cmake/build.py
```diff
@@ -313,16 +314,30 @@ def _get_make_arguments(self):
...
             cmd += ['--config', self._get_configuration(args)]
         else:
+            # Add make/ninja arguments for the number of job threads to use.
             job_args = self._get_make_arguments()
             if job_args:
+                # Add arguments for the number of job threads to use
+                # Arguably, _get_make_arguments() should be be configured to
+                # return arguments based on the current generator, not just for
+                # make (and ninja by coincidence). The '--' extension may be
+                # better returned from _get_make_arguments()
```
dirk-thomas: Please remove those comments which have been removed from the other PR.

KazNX: Will do. Just catching up on various comments.

colcon_cmake/task/cmake/__init__.py
KazNX: Good to know.

colcon_cmake/task/cmake/build.py
KazNX: Do you mean failure to parse - i.e, checking for a version number result which indicates failure -  or for failure of the await directive?

dirk-thomas: Your custom function might return `(0, 0, 0, 'failed')`.

colcon_cmake/task/cmake/__init__.py
KazNX: Right. In which case the following test will fail because the major part of the version is less than 3. At any rate, this is moot with the change to using pkg_resources for version parsing.

colcon_cmake/task/cmake/build.py
```diff
@@ -313,16 +315,26 @@ def _get_make_arguments(self):
...
-            CMAKE_EXECUTABLE, '--build', args.build_base,
-            '--target', 'install']
+        cmd = [CMAKE_EXECUTABLE]
+        cmake_ver = await get_cmake_version()
+        if cmake_ver and cmake_ver >= parse_version('3.15.0'):
+            # CMake 3.15+ supports invoking `cmake --install [--config]
+            # This only installs, whereas --build <dir> --target install will
+            # build (again) then install.
+            cmd += ['--install', args.build_base]
+        else:
```
dirk-thomas: Why is choosing the `else` block in case the version couldn't be retrieved the right choice?

KazNX: `--install` is a newer feature requiring 3.15 whereas the else statement uses `--build --target` which has had long term support. We can only guarantee support for `--install` if we can validate that the version number is `> 3.15`.

dirk-thomas: I understand that but are you expecting the get version function to fail for any CMake version? If not, please check for the potential returned error and not just silently use the else block.

colcon_cmake/task/cmake/build.py
KazNX: The return values have changed. This is no longer relevant.

dirk-thomas: While the way the version is parsed has changed the fact that the parsing can fail for an invalid string hasn't. And that case should be explicitly handled.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +210,29 @@ def get_visual_studio_version():
...
+
+    :returns: The version as parsed by the pkg_resources package
+    :rtype pkg_resources.extern.packaging.version.Version
+    """
+    output = await check_output([CMAKE_EXECUTABLE, '--version'])
+    lines = output.decode().splitlines()
+    ver_line = lines[0] if lines and len(lines) else None
+    if ver_line and len(ver_line):
+        # Extract just the version part of the string.
+        ver_re_str = r'^(?:cmake\s+version)?\s*(.*)$'
```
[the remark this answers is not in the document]
KazNX: Excellent question. I have no idea. If I don't remove the "cmake version" prefix, then the version number comes out as a LegacyVersion object, which will never pass the later checks for 3.15+. This means I have to do some parsing of the line here.

What do you think of using the following regular expression then: `r'^(?:.*)(\d+\.\d+\..*)'`. This reduces assumptions about the string. The requirements is that the pattern `#.#.<anything>` is present where `#` are numeric sequences.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +210,29 @@ def get_visual_studio_version():
...
+        ver_match = re.match(ver_re_str, ver_line, re.I)
+        if ver_match:
+            return parse_version(ver_match.group(1))
+    # Failed: return invalid version
+    return parse_version('0.0.0')
```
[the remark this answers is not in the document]
KazNX: Will fix.

colcon_cmake/task/cmake/__init__.py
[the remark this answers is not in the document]
dirk-thomas: What is the problem with comparing to a `LegacyVersion`?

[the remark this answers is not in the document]
KazNX: Running `cmake --verison` returns a string in the form: `cmake version 3.15.0`, optionally ending in `-rc1` for release candidates (at least in English). Simply reading the line as is as giving that to `pkg_resources.parse_version()` wil always generate a `LegacyVersion` because the string isn't just a version number. As I see it, `LegacyVersion` is not comparable; that is it is equivalent to version 0. I need to drop `cmake version ` in order to generate an informative version number object. The expression I propose has fewer assumptions about the cmake version string, assuming only that it ends in a `#.#.<text>` form. This allows any prefix to be on the version number line, regardless of language.

colcon_cmake/task/cmake/build.py
KazNX: Ok. I'll add an error message in the case wheret eh CMake version cannot be determined.

colcon_cmake/task/cmake/__init__.py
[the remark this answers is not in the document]
KazNX: Fix is m

colcon_cmake/task/cmake/__init__.py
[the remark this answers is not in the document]
KazNX: Note: resolving this issues will also resolve the other oustanding issues.

dirk-thomas:
Thanks for iterating on this. I have one more comment though: atm each package will determine the CMake version separately. It would be good if the function persists the determined version in a global variable and return that directly when being involved repeatedly.

dirk-thomas:
@KazNX Are you planning on addressing the last comment (determining the CMake version only once)?

KazNX:
@dirk-thomas Sorry, I've been busy on deliverables. Can you propose the best way to determine the CMake globally?

dirk-thomas:
> Can you propose the best way to determine the CMake globally?

@KazNX Please see my previous comment for that:

> It would be good if the function persists the determined version in a global variable and return that directly when being involved repeatedly.

See https://github.com/colcon/colcon-ros/blob/ba7472246ec4235fcae0ce71b055aff07bfec4c4/colcon_ros/task/cmake/__init__.py#L33-L59 for an example.

dirk-thomas:
@KazNX Friendly ping.

KazNX:
I have a caching solution in place now.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +210,55 @@ def get_visual_studio_version():
...
+        ver_re_str = r'^(?:.*)(\d+\.\d+\..*)'
+        ver_match = re.match(ver_re_str, ver_line, re.I)
+        if ver_match:
+            return parse_version(ver_match.group(1))
+    # Failed to parse version number.
+    return None
+
+
+# Global variable for the cached CMake version number.
+_cached_cmake_version = _parse_cmake_version()
```
dirk-thomas: Currently this determines the CMake version at import time of this module. Please change it to be lazily computed on the first invocation of `get_cmake_version` (as in the previously referenced example).

KazNX: Now lazy.

BTW, should I be clicking `Resolve conversation` or should you?

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +210,72 @@ def get_visual_studio_version():
...
+    cache the CMake version number. External API users should use
+    get_cmake_version()
+
+    :returns: The version as parsed by the pkg_resources package
+    :rtype pkg_resources.extern.packaging.version.Version
+    """
+    try:
+        output = subprocess.check_output([CMAKE_EXECUTABLE, '--version'])
+        lines = output.decode().splitlines()
+        ver_line = lines[0] if lines and len(lines) else None
```
dirk-thomas: The condition `len(lines)` is redundant with `lines`.

The logic can also be reduced by removing this line, changing the condition in the next line to just `if lines:`, and then below use `lines[0]` instead of `ver_line`.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +210,72 @@ def get_visual_studio_version():
...
+_cached_cmake_version = None
+
+
+def get_cmake_version():
+    """
+    Get the cached CMake version number if available and successfully parsed.
+
+    The result is None when a version number is not available or cannot be
+    parsed. This may occur if CMake is not present, is a very old version or is
+    a newer version where the version number string output has changed.
```
dirk-thomas: For what never version is that the case? Then the `--install` option wouldn't be used.

KazNX: I'm identifying the behaviour should the cmake version string format change in future. More of a maintenance note.

KazNX: I've modified the comment slightly.

colcon_cmake/task/cmake/__init__.py
KazNX: Fixed

KazNX:
All issues should be addressed now.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +211,64 @@ def get_visual_studio_version():
...
+        output = subprocess.check_output(
+            [CMAKE_EXECUTABLE, '--version'], stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
+        print('Failed to determine CMake version: ' + e.output.decode(),
+              file=sys.stderr)
+    else:
+        lines = output.decode().splitlines()
+        if lines:
+            # Extract just the version part of the string.
+            ver_re_str = r'^(?:.*)(\d+\.\d+\..*)'
```
dirk-thomas: I am not sure I understand why this regex was chosen. Can you please elaborate. It would be good to add a unit test to ensure this function does what it is supposed to do.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +211,64 @@ def get_visual_studio_version():
...
+            ver_match = re.match(ver_re_str, lines[0], re.I)
```
dirk-thomas: Why use `re.I` here? The pattern doesn't seem to contain any alphabetical characters.

colcon_cmake/task/cmake/__init__.py
KazNX: It does look odd starting with a non capturing group. I think this is also a throw back to when I had "cmake version" as part of the expression. Really I'm looking for `#.#.anything` where `#` is a number. CMake version numbers will often end with `.1rc2` or the like, so the patch number can't be read purely numerically.

dirk-thomas: > CMake version numbers will often end with .1rc2 or the like, so the patch number can't be read purely numerically.

I still don't understand why the regex would end in `\d+\..*`? In your example there is no dot between `.1` and `rc2`.

dirk-thomas: @KazNX Friendly ping.

colcon_cmake/task/cmake/__init__.py
dirk-thomas: @KazNX Are planning to update the patch accordingly?

dirk-thomas:
Also trying this patch with a CMake version new enough to trigger `--install` the invocation fails for me on Linux since it passes the `job_args` which seems to be not supported for the install option.

dirk-thomas:
@KazNX Friendly ping.

KazNX:
@dirk-thomas I've been waiting for responses on my comments. I'll have a look at the linux issue and streamlining the regex. Is there a simple example I can follow for the unit test?

dirk-thomas:
> I've been waiting for responses on my comments.

Can you please clarify what comments you are referring to?

As far as I can see I was the last person commenting on all pending conversation threads:
* regex pattern: https://github.com/colcon/colcon-cmake/pull/57#discussion_r364570229
* regex flag: https://github.com/colcon/colcon-cmake/pull/57#discussion_r367072807
* failing on Linux (and likely macOS): https://github.com/colcon/colcon-cmake/pull/57#issuecomment-574830801

> Is there a simple example I can follow for the unit test?

I am not sure what you are referring to? Regarding the failing build you should just try to build any CMake package using your patch.

KazNX:
>> I've been waiting for responses on my comments.

> Can you please clarify what comments you are referring to?

Looks like this is an unfamiliarity with GitHub on my part. I need to hit submit.


- regex parttern simplified
- regex flag removed
- Linux job_args removed

>> Is there a simple example I can follow for the unit test?

> I am not sure what you are referring to? Regarding the failing build you should just try to build any CMake package using your patch.

I was after an example to follow as I'm unsure how you are invoking unit tests. I added `test/test_parse_cmake_version.py` and it is failing the build because it can't import `colcon_cmake`. I'm not familiar with the cause.

colcon_cmake/task/cmake/build.py
```diff
@@ -333,10 +333,8 @@ def _get_make_arguments(self):
         if multi_configuration_generator:
             cmd += ['--config', self._get_configuration(args)]
         else:
-            # Add make/ninja arguments for the number of job threads to use.
```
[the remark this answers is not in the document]
dirk-thomas: Looking at the diff of this PR that doesn't seem to be the case.

dirk-thomas:
> `^(?:.*)` is questionable now. It's a non-capturing group. Better to replace with `.*`.

`.*(\d+\.\d+)` isn't correct since for an input string like `foo 12.3` the captured group will be only `2.3` because the leading wildcard is greedy. How about checking for the whitespace before the digit?

> `(\d+\.\d+\..*)` matches `(#.#.<anything>)` to match `major.minor.patch+comment`

The trailing `.*` matches anything. Could there be a "+comment" which you can't parse as a version? Why not limit the capture group to end after the digits?

> I'm unsure how you are invoking unit tests.

Just invoking `pytest` - the coverage flags are optional, see https://github.com/colcon/colcon-cmake/blob/c3e809cd57e38d10c15fa80df7fe0876252956a1/.travis.yml#L14

> I added `test/test_parse_cmake_version.py` and it is failing the build because it can't import `colcon_cmake`. I'm not familiar with the cause.

Locally the linters fail with the following:

```
test/test_copyright_license.py:10
In some files no copyright / license line was found
```

as well as 

```
1     W292 no newline at end of file
1     E302 expected 2 blank lines, found 1
1     E305 expected 2 blank lines after class or function definition, found 1
1     E501 line too long (80 > 79 characters)
2     F401 'pkg_resources.parse_version' imported but unused
1     Q000 Remove bad quotes
```

---

Regarding the comments: lets first focus on getting the functional logic of the patch right and working as well as the tests turn over. Then we can revisit the comments and other nitpicks.

KazNX:
I've updated the regular expression to be less greedy with leading characters. I'm also ignoring any non-numeric suffix. It turns out CMake can have many suffixes including '-dirty' on working versions. I don't see it being worthwhile to distinguish these. I've tried a number of expected and some unexpected strings in `test_parse_cmake_version.py` including checking the greedy case you've highlighted.

I've fixed the linter issues. I didn't notice these because it's the following error which has my attention:

> ImportError while importing test module '/home/travis/build/colcon/colcon-cmake/test/test_parse_cmake_version.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> test/test_parse_cmake_version.py:4: in <module>
>     from colcon_cmake.task.cmake import parse_cmake_version_string
> E   ImportError: No module named 'colcon_cmake'

Clearly it's failing to import the module I'm trying to test. This is where I'm unsure of why though. I can run the test from the command line. By contrast, colcon-core tests do something similar, importing modules from within colcon-core.

dirk-thomas:
> Clearly it's failing to import the module I'm trying to test. This is where I'm unsure of why though. I can run the test from the command line. By contrast, colcon-core tests do something similar, importing modules from within colcon-core.

Please try to create an empty file `test/__init__.py`.

KazNX:
That seems to have done the trick. Thanks. That should address all your oustanding issues.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -215,24 +215,22 @@ def get_visual_studio_version():
...
+It may also have a boolean value False when the CMake version could not be
+determined to avoid trying to determine it again.
 """
 _cached_cmake_version = None
 
 
 def get_cmake_version():
     """
-    Get the cached CMake version number if available and successfully parsed.
+    Get the CMake version.
```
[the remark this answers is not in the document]
dirk-thomas: I agree. Currently the additional information if in the following paragraph. Please feel free to suggest a better first sentence (which passes the linter in terms of line length).

colcon_cmake/task/cmake/__init__.py
```diff
@@ -262,23 +260,9 @@ def get_cmake_version():
...
-    Additional text may follow, however only the first line is parsed.
-
-    Parsing is performed by the pkg_resources package and the returned object
-    is a Version object from that package.
-
-    This function blocks on the execution of cmake and should only be used to
-    cache the CMake version number.
-
-    External API users should use get_cmake_version()
+    Parse the CMake version printed by `CMAKE_EXECUTABLE --version`.
```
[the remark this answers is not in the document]
dirk-thomas: Imo some of the sentences where not helpful, e.g.:

> This function blocks on the execution of cmake and should only be used to cache the CMake version number.

Since the function is not `async` it clearly is blocking. I don't see why that should be part of the docblock.

---

> Parsing is performed by the pkg_resources package and the returned object is a Version object from that package.

The return type clearly documents how the version is represented.

---

> The version number is parse from the output of 'cmake --version' and is expected in the form 'cmake version #.#.#' where each '#' character represents part of the version number.
> Additional text may follow, however only the first line is parsed.

The part - that the output of `cmake --version` is parsed - is imo the important part for the reader. The other text isn't relevant to the user of the API - at most it could be a comment documenting how the logic works if a developer wants to read / understand the implementation of the function.

colcon_cmake/task/cmake/build.py
```diff
@@ -325,13 +325,10 @@ def _get_make_arguments(self):
...
         else:
             # Fallback to building the install target.
-            #
-            # Log an error if we are here because we failed to get a meaningful
-            # version number rather than simply using an unsupported version.
             if not cmake_ver:
-                logger.error(
-                    'Unable to determin CMake version. Fallback to building '
-                    'the install target in place of cmake --install.')
+                logger.warning(
```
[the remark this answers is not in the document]
dirk-thomas: Can you please clarify why `not cmake_ver` would be `True` and this warning be triggered if a project has no install target?

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +211,81 @@ def get_visual_studio_version():
...
+        cmake_version = _parse_cmake_version()
+        if cmake_version is not None:
+            _cached_cmake_version = cmake_version
+        else:
+            # flag to avoid retrying on subsequent calls
+            _cached_cmake_version = False
+
+    if _cached_cmake_version is False:
+        # in case the CMake version could not be determined
+        return None
```
dirk-thomas: The logic around the cache value is pretty convoluted coming from the fact that the function uses a different "negative" value (`None`) than the cached value (`False`).

I aligned those in 6390dd148c23f910309afe802c76cbdc2dddff3d which gets rid of the majority of the logic in this function.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +211,71 @@ def get_visual_studio_version():
...
+    if _cached_cmake_version is False:
+        _cached_cmake_version = _parse_cmake_version()
+    return _cached_cmake_version
+
+
+def parse_cmake_version_string(version_string):
+    """
+    Parse the given CMake version string.
+
+    This is for internal use and testing.
```
dirk-thomas: Since the function is meant to be internal I removed this line from the docblock and instead prefixed the function name with an underscore: 631e008b53811e6031dc42cb30d3d1b6db27cf15.

colcon_cmake/task/cmake/build.py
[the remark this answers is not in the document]
KazNX: I will be `False` on failure to parse rather than `None`. This is so the upstream doce can distuguish between no parsing attempt made (`None`) and failed to parse (`False`).

test/test_parse_cmake_version.py
```diff
@@ -0,0 +1,44 @@
...
+        parsed_version = _parse_cmake_version_string(test_item[0])
+        expected_ver = test_item[1]
+        # print(test_item[0], ':', parsed_version, 'expected:', expected_ver)
+        if expected_ver is None:
+            # Input string was garbage. Assert parsing failed.
+            assert parsed_version is None
+        else:
+            assert parsed_version._version.release[0] == expected_ver[0]
+            assert parsed_version._version.release[1] == expected_ver[1]
+            assert parsed_version._version.release[2] == expected_ver[2]
```
dirk-thomas: Simplified the test logic in 83bd8e76dc6e4dabca5fecbdb48f7f3933658a14.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -208,3 +211,69 @@ def get_visual_studio_version():
...
+    Parse the given CMake version string.
+
+    Expects strings of the form 'cmake version 3.15.1'.
+
+    :param str version_string: The version string to parse.
+    :returns: The parsed version string or None on failure to parse.
+    :rtype pkg_resources.extern.packaging.version.Version
+    """
+    # Extract just the version part of the string.
+    ver_re_str = r'^(?:.*\s)?(\d+\.\d+\.\d+).*'
```
dirk-thomas: Using `\s` for matching whitespace: see 83bd8e76dc6e4dabca5fecbdb48f7f3933658a14.

test/test_parse_cmake_version.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2016-2018 Dirk Thomas
```
dirk-thomas: Please consider use your name for the copyright of this file since you wrote it. Also the year should 2020.

KazNX: Fixed

colcon_cmake/task/cmake/__init__.py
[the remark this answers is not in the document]
KazNX: If I'm right with the line limit being 80 characters, then the orignal comment fit. Otherwise, how about:

> Query the CMake --version results if available.

Otherwise, I'm not going to deliberate over this one. I'd like to complete the PR.

colcon_cmake/task/cmake/__init__.py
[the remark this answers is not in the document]
KazNX: Ok. Again, I'm not going to hold the PR up on these details.

colcon_cmake/task/cmake/__init__.py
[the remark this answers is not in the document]
dirk-thomas: I wouldn't use "query". Also "results" doesn't sound right to me. And "if available" is fuzzy too.
How about:

> Get the version as reported by `CMAKE_EXECUTABLE --version` if parsable.

The redundant detail can then be removed from `:returns:`:

> The CMake version, or None when the version number could not be determined

dirk-thomas:
Some ROS 2 CI builds to double check:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/9172/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/7485/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/9088/)

KazNX:
> Some ROS 2 CI builds to double check:

They all look good, but are they running with this change or does something need to be done in order to test this PR with those builds?

dirk-thomas:
> are they running with this change

Yes, I have pushed the changes from this PR to a branch within this repo (https://github.com/colcon/colcon-cmake/tree/enhancement/use-cmake-install) and the jobs are using that branch (see CI parameter `colcon_branch`).

This is good to be merged. Thank you for this nice improvement. I am sorry for the multiple rounds of iteration and appreciate that you kept iterating on it. :bowing_man:

PR #62, opened 2019-12-20
Use `--cmake-concat-compile-commands` to concatenate multiple compile_commands.json files into a single file

### LINKED ISSUE #61: Option to concatenate multiple compile_commands.json files into a single file
When we run `colcon build` with the following `cmake-args` option, we can got `compile_commands.json` for each package.

```
colcon build --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
```

It's very useful to use auto complete of programming ROS / ROS 2 by the language server protocol of `clangd`.
It'll be even more useful to develop multiple packages at a time if the multiple `compile_commands.json` to unify one `compile_commands.json`.

I'm thinking about adding a `--cmake-concat-compile-commands` option. How do you think about it? I try to implement it soon.

Thanks to the `colcon` extensibility, my development experience is better than before!

youtalk:
Resolved #61

This PR is a draft PR to resolve #61 issue. You can try it by the following command.
```
$ colcon build --cmake-concat-compile-commands --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
$ less build/compile_commands.json
```

It is very useful for me but I know it has no test and the process is not efficient. 
After making this PR, I've noticed that I need to implement a new `colcon` extension/package instead of modifying the `colcon-cmake` package.

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -2,6 +2,7 @@
 # Licensed under the Apache License, Version 2.0
 
 import ast
+import json
 import os
 from pathlib import Path
 import re
@@ -105,6 +106,9 @@ class CmakeBuildTask(TaskExtensionPoint):
             args, env, additional_targets=additional_targets)
         if rc and rc.returncode:
             return rc.returncode
+
+        if args.cmake_concat_compile_commands:
+            self._concatenate_compile_commands(args)
 
         # skip install step if a specific target was requested
         if not args.cmake_target:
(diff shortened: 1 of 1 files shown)


galou:
Did you manage to implement the said `colcon` extension? Where?

youtalk:
Not yet... Are you interested in it?

galou:
Yes, because I'm using YouCompleteMe and the setup would be very easy with a single `compile_commands.json` that I would symlink into my workspace directory. Without this extension, I'd have to symlink each `compile_commands.json` into the source directory of each package.

Is there any documentation how to develop an extension for `colcon`, so that I could maybe help you.

youtalk:
I think there is no documentation yet but:
https://colcon.readthedocs.io/en/released/developer/contribution.html#new-packages-extensions

dirk-thomas:
Colcon simply uses Python enter points to register extensions (and even extension points).

The most common extension points are described in https://colcon.readthedocs.io/en/released/developer/extension-point.html. If you install `colcon-devtools` you can use the verbs `extension-points` and `extensions` to get complete lists.

The overall program flow of the relevant `build` command is visualized in https://colcon.readthedocs.io/en/released/developer/program-flow.html and might help to show the big picture.

Other than that there is always the code to browse or you can ask questions here.

galou:
@dirk-thomas: thanks for offering help! My first question would be: how do I register an extension so that colcon finds it? Without installing anyting, I ran `colcon build --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON --cmake-concat-compile-commands` to see where Python was looking for extra modules but it looks that every unknown argument is treated as the name of a source directory.

dirk-thomas:
> My first question would be: how do I register an extension so that colcon finds it?

My previous comment described exactly that.

I would suggest to take a look at the sources of `colcon-core` to see an example how it contributes commands line arguments.

PR #65, opened 2020-02-05
Respect ctest --build-config if explicitly provided.

KazNX:
Change ctest calls to respect `-C` or `--build-config` if explicitly given. Fallback to previous behaviour of reading `CMAKE_BUILD_TYPE` from `CMakeCache.txt`.

### DIFF
--- a/colcon_cmake/task/cmake/test.py
+++ b/colcon_cmake/task/cmake/test.py
@@ -66,14 +66,28 @@ class CmakeTestTask(TaskExtensionPoint):
 
         # CTest arguments
         ctest_args = [
-            # choose configuration, required for multi-configuration generators
-            '-C', self._get_configuration_from_cmake(args.build_base),
             # generate xml of test summary
             '-D', 'ExperimentalTest', '--no-compress-output',
             # show all test output
             '-V',
             '--force-new-ctest-process',
         ]
+
+        # Check for explicit '-C' or '--build-type' in args.ctest_args.
+        # If not, we'll add it based on the CMakeCache CMAKE_BUILD_TYPE value.
+        lookup_config = True
+        if args.ctest_args:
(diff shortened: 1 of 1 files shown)


colcon_cmake/task/cmake/test.py
```diff
@@ -66,14 +66,28 @@ def add_arguments(self, *, parser):  # noqa: D102
...
+        # Check for explicit '-C' or '--build-type' in args.ctest_args.
+        # If not, we'll add it based on the CMakeCache CMAKE_BUILD_TYPE value.
+        lookup_config = True
+        if args.ctest_args:
+            for arg in args.ctest_args:
+                if arg == '-C' or arg == '--build-config':
+                    lookup_config = False
+                    break
+
+        if lookup_config:
```
dirk-thomas: Please simplify this block of logic. A single condition should do just fine. I would suggest to first concatenate the above assignment of the static arguments above with the dynamic arguments passed by the user:

```
ctest_args = [
    # generate xml of test summary
    '-D', 'ExperimentalTest', '--no-compress-output',
    # show all test output
    '-V',
    '--force-new-ctest-process',
] + (args.ctest_args or [])
```

Then the logic to conditionally add `-C` becomes trivial:

```
if '-C' is not in ctest_args and '--build-config' is not in ctest_args:
    # choose configuration, required for multi-configuration generators
    ctest_args[0:0] = [
        '-C', self._get_configuration_from_cmake(args.build_base),
    ]
```

The choice to add the inferred config in the front is to keep the existing behavior as well as keep it separate from the user provided arguments at the end.

KazNX: Done.

I agree in keeping `-C` at the front. I don't spend enough time in python for the list splicing syntax to be second nature.

python magic!

PR #68, opened 2020-03-16
Expose '--ctest-pytest-with-coverage' option

christophebedard:
This exposes a `--ctest-pytest-with-coverage` option for `'cmake'` packages. The intent is to use this to generate coverage information for pytest tests with CMake.

Related to  https://github.com/ament/ament_cmake/pull/226

### DIFF
--- a/colcon_cmake/task/cmake/test.py
+++ b/colcon_cmake/task/cmake/test.py
@@ -30,6 +30,10 @@ class CmakeTestTask(TaskExtensionPoint):
             help='Pass arguments to CTest projects. '
             'Arguments matching other options must be prefixed by a space,\n'
             'e.g. --ctest-args " --help"')
+        parser.add_argument(
+            '--ctest-pytest-with-coverage',
+            action='store_true',
+            help='Generate coverage information for Python tests')
 
     async def test(self, *, additional_hooks=None):  # noqa: D102
         pkg = self.context.pkg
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Since this option is not applicable to all CMake packages but only ROS/ament CMake packages this should be added to `colcon-ros` instead.

Before updating this I would recommend to wait for the feedback on ament/ament_cmake#226 to settle. Maybe the best way would be to set the to-be-created CMake option if the existing argument [--pytest-with-coverage](https://github.com/colcon/colcon-core/blob/150255db5acdc242dba06e4cdeca359aee6b50f2/colcon_core/task/python/test/pytest.py#L44) is set?

christophebedard:
> Since this option is not applicable to all CMake packages but only ROS/ament CMake packages this should be added to `colcon-ros` instead.

Makes sense!

> Before updating this I would recommend to wait for the feedback on [ament/ament_cmake#226](https://github.com/ament/ament_cmake/pull/226) to settle.

Will do.

> Maybe the best way would be to set the to-be-created CMake option if the existing argument [--pytest-with-coverage](https://github.com/colcon/colcon-core/blob/150255db5acdc242dba06e4cdeca359aee6b50f2/colcon_core/task/python/test/pytest.py#L44) is set?

Correct me if I'm wrong, but what you're suggesting over in https://github.com/ament/ament_cmake/pull/226#issuecomment-604219217 is to use a CMake `option(...)`, which needs to be set at build time, while that existing `--pytest-with-coverage` flag is used when testing (`colcon test`).

If the above is correct, yes, a similar flag for `colcon build` makes sense.

christophebedard:
I'm going to close this in favour of https://github.com/colcon/colcon-ros/pull/101. Feel free to keep the discussion here or just move it there.

Issue #70, opened 2020-03-24
Improve help message for --ctest-args


rotu:
`colcon test --packages-select rosbag2_transport --ctest-args " --help"` is expected to show help for ctest arguments but does not. Output is similar to if tests were run and succeeded.

```
colcon test --packages-select rosbag2_transport --ctest-args " --help"
Starting >>> rosbag2_transport
Finished <<< rosbag2_transport [0.19s]          

Summary: 1 package finished [1.06s]
```

The output from `ctest --help` can be found in `log/latest_test/rosbag2_transport/stdout.log`.

Perhaps the help text should come out in stdout or the help for the `--ctest-args` option should use an argument other than " --help" as in:

```
$ colcon test --help
usage: colcon test [-h] [--build-base BUILD_BASE] [--install-base INSTALL_BASE] [--merge-install]
...
Arguments for 'cmake' packages:
  --ctest-args [* [* ...]]
                        Pass arguments to CTest projects. Arguments matching other options must be prefixed by a space,
                        e.g. --ctest-args " --help"
```

dirk-thomas:
Since `colcon` doesn't show `stdout` output by default and `ctest --help` prints to `stdout` (not `stderr`) this is expected behavior. You can pass e.g. `--event-handlers console_direct+` to see the `stdout` output of the package.

rotu:
I understand *why* it happens. Perhaps, since the `—help` option has such caveats, it would be better to use a different example in the help text.

dirk-thomas:
> ... use a different example in the help text.

Please include the relevant context in you original ticket. It is very difficult to follow what you are referring to if you don't. In this case I assume you are referring to the help message of the `--ctest-args` option: https://github.com/colcon/colcon-cmake/blob/fff16c206886a63517c848792d3fb68601c4a8cd/colcon_cmake/task/cmake/build.py#L39-L41

rotu:
Yep. I wrote this ticket in a pretty scattered way, but I think you caught the gist. If I were to try to lay out my thought process, here it is:

1. I’d like to see how to use `colcon` effectively for running a single test case. In my case, I knew the test was kicked off by `ctest`. So I use `colcon test —help` to see what options might help me.
2. The relevant argument for test selection is `—ctest-args` so let’s try the example it gives.
3. Oops that didn’t seem to do anything...

My hangups here:
[...]

PR #72, opened 2020-03-25
add note to help message that stdout might not be shown by default

dirk-thomas:
Similar to colcon/colcon-core#328. Partially addresses colcon/colcon-cmake#70.


dirk-thomas:
> This nitpick from the `colcon-core` PR applies here, too: colcon/colcon-core#328 (comment)

Addressed in d130f2be443c01c2ccc98ba72840aee5a3a8a5cb.

PR #73, opened 2020-03-27
rename rc -> completed

dirk-thomas:
Similar to colcon/colcon-core#235.

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -84,8 +84,8 @@ class CmakeBuildTask(TaskExtensionPoint):
             environment_callback(env)
 
         rc = await self._reconfigure(args, env)
-        if rc and rc.returncode:
-            return rc.returncode
+        if rc:
+            return rc
 
         # ensure that CMake cache contains the project name
         project_name = get_variable_from_cmake_cache(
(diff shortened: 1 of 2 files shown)


colcon_cmake/task/cmake/build.py
```diff
@@ -101,15 +101,15 @@ def add_arguments(self, *, parser):  # noqa: D102
 
         rc = await self._build(
             args, env, additional_targets=additional_targets)
-        if rc and rc.returncode:
-            return rc.returncode
+        if rc:
+            return rc
```
mjcarroll: Is this supposed to be `rc.returncode` to keep the return type consistent?

colcon_cmake/task/cmake/build.py
```diff
@@ -236,10 +237,10 @@ def _get_last_cmake_args_path(self, build_base):
                 job_args = self._get_make_arguments()
                 if job_args:
                     cmd += ['--'] + job_args
-            rc = await check_call(
+            completed = await check_call(
                 self.context, cmd, cwd=args.build_base, env=env)
-            if rc and rc.returncode:
-                return rc
+            if completed.returncode:
+                return completed.returncode
```
mjcarroll: This is also changing the return type.

dirk-thomas: This change goes together with your other comment. The `_build` method now returns just the integer return code (instead of a `CompletedProcess` before).

colcon_cmake/task/cmake/build.py
dirk-thomas: This change goes together with your other comment. The `_build` method now returns just the integer return code (instead of a `CompletedProcess` before).

PR #74, opened 2020-03-28
add dependency on cmake in Debian package

mikaelarguedas:
After installing `colcon-core` from apt, I am unable to build a CMake package despite having `colcon-cmake` installed. For convenience it would be good to have an explicit dependency on `cmake` for the debs of colcon-cmake

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,5 +1,5 @@
 [colcon-cmake]
 No-Python2:
-Depends3: python3-colcon-core, python3-colcon-library-path, python3-colcon-test-result (>= 0.3.3)
+Depends3: cmake, python3-colcon-core, python3-colcon-library-path, python3-colcon-test-result (>= 0.3.3)
 Suite: xenial bionic focal stretch buster
 X-Python3-Version: >= 3.5


mikaelarguedas:
> E.g. in ROS packages we commonly do not declare a dependency on cmake or build-essential. Instead we let the user install what he wants in order to be able to build a package (independent of the build tool).

This is fair in the context of packages that don't *need* CMake at runtime. However I don't think this should apply to packages that invoke the `cmake` executable directly and cannot run without it (from my understanding). These packages do *require* cmake.
It seems more conventional to always install the *required* packages and to let user opt out of the recommended packages if they want to (which can be easily done using the `--no-install-recommends` apt flag).

The fact that `cmake` recommends `gcc` and `make` in Ubuntu is something one can't control, but it seems unconventional to decide not to declare a required dependency based on that information.

mjcarroll:
I think that it makes sense that `colcon-cmake` depends on `cmake`, and the exclusion with `--no-install-recommends` would prevent from getting other things installed.

My only reservation here is that it if this is done consistently across colcon packages (eg should `colcon-zsh` depend on `zsh`?  should `colcon-bazel` depend on `bazel`?) it will make `colcon-common-extensions` quite large.

mikaelarguedas:
From my perspective these are slightly different scenarios:

- `colcon-zsh` generates `zsh` files but doesnt need zsh to be installed to do so, so it is not a required dependency (same would apply to other shells)
- `colcon-bazel` would not be able to run without bazel installed, so bazel is a required dep (though is not part of colcon-common-extensions)

>  it will make colcon-common-extensions quite large.

A closer look at which packages in `colcon-common-extensions` would bring in extra dependencies is needed to answer that. At a glance apart from `colcon-cmake` the other packages seem to already bring in their dependencies (e.g. `colcon-ros` brings in `catkin_pkg` etc)

mjcarroll:
That's fair, for some reason I thought that `colcon-bazel` was in the common extensions, but it looks like I was mistaken.  In that case, I think that this change makes sense (along with a corresponding change in `colcon-bazel`).

I also agree with your logic around `zsh`/`powershell`/etc

dirk-thomas:
Ok, sounds good to me then.

> along with a corresponding change in `colcon-bazel`

It would be great if you could create a corresponding pull request for `colcon-bazel` then.

mikaelarguedas:
> It would be great if you could create a corresponding pull request for colcon-bazel then.

I can't find bazel in the official ubuntu repositories. Maybe it's only available via ppa

mjcarroll:
> Maybe it's only available via ppa

That is exactly the case as of `bionic`: https://docs.bazel.build/versions/master/install-ubuntu.html

PR #77, opened 2020-04-13
Exclude test directory from package search

cottsay:
The files in the `test` directory are getting installed as an additional package.

The `__init__.py` was added in #57, so this bug has existed since 0.2.18.


dirk-thomas:
I tried to remove the `test/__init__.py` file instead: c280d5b037dbef6fdd0865d25e44f89683747376. To make that work Travis CI needs to invoke `python -m pytest` instead of `pytest`: 4f2a86feb14e43f27c28035100150c7152c2834e.

Since that seems to work I would rather go that route. Does that make sense to you?

cottsay:
That's how I'm [invoking the tests already](https://src.fedoraproject.org/rpms/python-colcon-cmake/blob/master/f/python-colcon-cmake.spec#_48), so it works for me. Thanks!

dirk-thomas:
Thanks for the fix.

PR #78, opened 2020-04-13
Fix crash under Python3.9

rotu:
As per docs:

https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getchildren
```
getchildren()
    Deprecated since version 3.2, will be removed in version 3.9: Use list(elem) or iteration.
```

### DIFF
--- a/colcon_cmake/test_result/ctest.py
+++ b/colcon_cmake/test_result/ctest.py
@@ -59,15 +59,13 @@ class CtestTestResult(TestResultExtensionPoint):
                 continue
 
             # look for a single 'Testing' child tag
-            try:
-                [testing] = root
-            except ValueError:
+            children = list(root)
+            if len(children) != 1:
                 logger.warning(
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for reporting this. Based on the referenced documentation wouldn't it be sufficient to replace the following line:

https://github.com/colcon/colcon-cmake/blob/8a2ddbdb90dd5542f0f9bf10e7b89ccf6c7f48b3/colcon_cmake/test_result/ctest.py#L62

with:

    children = list(root)

rotu:
Yes, it would. I find this way is more idiomatic and it removes the repetition of "`children[0]`". You can think of it as `children = list(root); testing = children[0]`.

dirk-thomas:
I rather keep the diff minimal (7922b617655884ae9538b68291e6eebf863d8d86). I don't think the repetition of `children[0]` is worth introducing another variable.

PR #85, opened 2020-06-22
Handle cmake '-G<generator>' syntax

KazNX:
CMake supports generators specified using '-G <generator>' (2 arguments) or as '-G<generator>' (1 argument). This change adds colcon support for the latter, one argument syntax.


dirk-thomas:
Thanks for the patch. I refactored the logic in d82515d5ecaf9dad8466ca15de301ef2b7cf25af to avoid the extra nesting and avoid touching the existing line.

Please use the GitHub specific keyword "Fixes #86" in the description in the future to automatically close the referenced issue.

KazNX:
Noted. I created these in the opposite order :)

KazNX:
On the logic front, I was avoiding specifying the string `-G` twice

PR #88, opened 2020-07-22
only generate workspace-level compile_commands.json when CMake was invoked

### LINKED ISSUE #81: Workspace-level "compile_commands.json" is being generated by "colcon test"
It looks like the new `CompileCommandsEventHandler` (introduced by https://github.com/colcon/colcon-cmake/pull/69) is being invoked by `colcon test`.

For example, if you do
```
colcon build --event-handlers compile_commands- --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
colcon test  # don't specify any --event-handlers options
```
there will still end up being a workspace-level `compile_commands.json` under `build/`.

Is this intended behavior? I think it's more intuitive if the workspace-level`compile_commands.json` is only generated during `colcon build`.

dirk-thomas:
Fixes #81.

It not only skips the (re-)generation of the workspace-level `compile_commands.json` file when `colcon test` is invoked as described in the referenced ticket but also skips it when e.g. only non-CMake packages are being built.

### DIFF
--- a/colcon_cmake/event_handler/compile_commands.py
+++ b/colcon_cmake/event_handler/compile_commands.py
@@ -4,6 +4,9 @@
 import os
 from pathlib import Path
 
+from colcon_cmake.task.cmake import CMAKE_EXECUTABLE
+
+from colcon_core.event.command import Command
 from colcon_core.event.job import JobQueued
 from colcon_core.event.job import JobUnselected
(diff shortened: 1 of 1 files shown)


mikepurvis:
This does the thing for me now, sans workaround. Thanks!

PR #89, opened 2020-08-10
Copy CTest results to custom location

### LINKED ISSUE #79: CTest result files are not in --test-result-base
Example:
```
colcon build --test-result-base test_results
colcon test --test-result-base test_results
colcon test-result --test-result-base test_results --all --result-files-only
```
Expected behavior would be that the CTest `Test.xml` results from any ctest invocations would be printed. However, those files ended up in the `--build-base`. This seems to contradict the explicit purpose of `--test-result-base`, "the base path for all test results".

From the example above, if you run `colcon test-result --all --result-files-only` so that `--test-result-base` defaults to be the same as `--build-base`, the paths to the CTest `Test.xml` results are printed as expected.

dirk-thomas:
Fixes #79.

### DIFF
--- a/colcon_cmake/task/cmake/test.py
+++ b/colcon_cmake/task/cmake/test.py
@@ -2,6 +2,9 @@
 # Licensed under the Apache License, Version 2.0
 
 import os
+import shutil
+from contextlib import suppress
+from pathlib import Path
 
 from colcon_cmake.task.cmake import CTEST_EXECUTABLE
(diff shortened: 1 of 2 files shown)


dirk-thomas:
> Pretty sure the `build_base` gets deleted at the end of the job, but before the test results are inspected by Jenkins.

Good point. I didn't think of that use case. I modified the patch to do a plain copy: 23005763aa7a5b62977ea1b87bd6c437506d2255.

PR #90, opened 2020-08-20
Remove configuration sniffing

rotu:
colcon test currently can run with the wrong build type. If `CMAKE_BUILD_TYPE` is unset, the cache line will be `CMAKE_BUILD_TYPE:STRING=`, which this logic previously assumed means 'Release'. That is not a safe assumption, and in multi-config setups the configuration may be influenced by `CMAKE_DEFAULT_BUILD_TYPE`, `CMAKE_DEFAULT_CONFIGS`, `CTEST_CONFIGURATION_TYPE`, or `DEFAULT_CTEST_CONFIGURATION_TYPE`.

### DIFF
--- a/colcon_cmake/task/cmake/test.py
+++ b/colcon_cmake/task/cmake/test.py
@@ -77,14 +77,6 @@ class CmakeTestTask(TaskExtensionPoint):
             '--force-new-ctest-process',
         ] + (args.ctest_args or [])
 
-        # Check for explicit '-C' or '--build-type' in args.ctest_args.
-        # If not, we'll add it based on the CMakeCache CMAKE_BUILD_TYPE value.
-        if '-C' not in ctest_args and '--build-config' not in ctest_args:
-            # choose configuration, required for multi-configuration generators
-            ctest_args[0:0] = [
-                '-C', self._get_configuration_from_cmake(args.build_base),
-            ]
-
         if args.retest_until_fail:
             count = args.retest_until_fail + 1
             ctest_args += [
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Have you verified that this change still works on Windows with Visual Studio which is the case why this logic was added in the first place?

rotu:
Working on it. How is CI normally run against ROS2 for colcon PRs?

dirk-thomas:
> How is CI normally run against ROS2 for colcon PRs?

The jobs on ci.ros2.org have a parameter `CI_COLCON_BRANCH` to pass a branch name in the `colcon` repositories. Specifying a fork isn't possible atm though.

rotu:
Running CI here: https://github.com/rotu/colcon-cmake/actions/runs/222969761

dirk-thomas:
> Running CI here: https://github.com/rotu/colcon-cmake/actions/runs/222969761

@rotu The referenced build failed. Any update on this?

rotu:
Unfortunately, no. I recommend using your regular CI --- it's clear that my attempt at using GitHub CI for this is a little half-baked.

dirk-thomas:
Please see https://ci.ros2.org/job/ci_windows/12408/ which doesn't specify a build type. The test invocations fail since `-C` is not specified anymore:

> Test not available without configuration.  (Missing "-C <config>"?)

dirk-thomas:
I will go ahead and close this ticket for now since the approach doesn't work. Please feel free to iterate on this to support Ninja better. Either the PR can be reopened when the branch was revised or a new PR can be opened.

PR #92, opened 2020-08-27
use msbuild -version, not %VisualStudioVersion%

rotu:
When building on Windows, instead of using `%VisualStudioVersion%`, use `msbuild -version`

This is intended to make `colcon-cmake` more friendly to Github actions. The `microsoft/setup-msbuild` action does not populate `%VisualStudioVersion%`, so `colcon-cmake` crashes on that platform.

Here is a workflow demonstrating the issue and the proposed fix:

https://github.com/rotu/ros2-ci/actions/runs/233271645/workflow

### DIFF
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -6,6 +6,7 @@ import re
 import shutil
 import subprocess
 import sys
+import functools
 
 from colcon_core.environment_variable import EnvironmentVariable
 from colcon_core.subprocess import check_output
@@ -206,6 +207,7 @@ def get_project_file(path, target):
     return project_file
 
 
+@functools.lru_cache(maxsize=1)
 def get_msbuild_version():
     """
     Get the version of msbuild or throw an exception.
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -214,10 +214,9 @@ def get_msbuild_version():
     """
     cp = subprocess.run(
         ['msbuild', '-version', '-noLogo'],
-        capture_output=True,
-        encoding=sys.getdefaultencoding(),
+        capture_output=True
     )
-    return cp.stdout
+    return cp.stdout.decode(encoding=sys.getdefaultencoding())
 
 
 def get_visual_studio_version():
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -206,6 +206,19 @@ def get_project_file(path, target):
     return project_file
 
 
+def get_msbuild_version():
+    """
+    Get the version of msbuild or throw an exception.
+
+    :rtype: str
+    """
+    cp = subprocess.run(
+        ['msbuild', '-version', '-noLogo'],
+        capture_output=True
+    )
+    return cp.stdout
+
+
 def get_visual_studio_version():
     """
     Get the Visual Studio version.
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -214,7 +214,8 @@ def get_msbuild_version():
     """
     cp = subprocess.run(
         ['msbuild', '-version', '-noLogo'],
-        capture_output=True
+        capture_output=True,
+        encoding=sys.getdefaultencoding(),
     )
     return cp.stdout
 
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -214,11 +214,11 @@ def get_msbuild_version():
 
     :rtype: str
     """
-    cp = subprocess.run(
+    completed_process = subprocess.run(
         ['msbuild', '-version', '-noLogo'],
         capture_output=True
     )
-    return cp.stdout.decode(encoding=sys.getdefaultencoding())
+    return completed_process.stdout.decode(encoding=sys.getdefaultencoding())
 
 
 def get_visual_studio_version():
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -154,16 +154,16 @@ class CmakeBuildTask(TaskExtensionPoint):
                 raise RuntimeError(
                     'msbuild is not found, '
                     'please run within a Visual Studio Command Prompt.') from e
-            major_version = msbuild_version.split('.')[0]
-            if major_version not in ('14', '15', '16'):
+            major_version = int(msbuild_version.split('.')[0])
+            if major_version < 14:
                 raise RuntimeError(
                     "Unknown / unsupported VS version '{msbuild_version}'"
                     .format_map(locals()))
-            cmake_args += ['-G', 'Visual Studio ' + major_version]
+            cmake_args += ['-G', 'Visual Studio ' + str(major_version)]
             # choose 'x64' on VS 14 and 15 if not specified explicitly
             # since otherwise 'Win32' is the default for those
             # newer versions default to the host architecture
-            if '-A' not in cmake_args and major_version in ('14', '15'):
+            if '-A' not in cmake_args and major_version < 16:
                 cmake_args += ['-A', 'x64']
         if CMAKE_EXECUTABLE is None:
             raise RuntimeError("Could not find 'cmake' executable")
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -162,7 +162,8 @@ class CmakeBuildTask(TaskExtensionPoint):
                     .format_map(locals())) from e
             if major_version < 14:
                 raise RuntimeError(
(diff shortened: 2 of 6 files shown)


rotu:
FYI, the stuff in `test.py` and `__init__.py` was changed because the linter complained about it. I can revert those changes but it will make the tests fail locally for me

```
(.venv) C:\Users\dan\Documents\colcon-ws>pip list
Package               Version Location
--------------------- ------- -------------------------------------------------------
astroid               2.4.2
atomicwrites          1.4.0
attrs                 20.1.0
colcon-bash           0.4.2   c:\users\dan\documents\colcon-ws\.venv\src\colcon-bash
colcon-cmake          0.2.25  c:\users\dan\documents\colcon-ws\.venv\src\colcon-cmake
colcon-core           0.6.0   c:\users\dan\documents\colcon-ws\.venv\src\colcon-core
colcon-library-path   0.2.1
colcon-test-result    0.3.8
colorama              0.4.3
coloredlogs           14.0
coverage              5.2.1
distlib               0.3.1
empy                  3.3.4
flake8                3.8.3
flake8-blind-except   0.1.1
flake8-builtins       1.5.3
flake8-class-newline  1.6.0
flake8-comprehensions 3.2.3
flake8-deprecated     1.3
flake8-docstrings     1.5.0
flake8-import-order   0.18.1
flake8-polyfill       1.0.2
flake8-quotes         3.2.0
humanfriendly         8.2
iniconfig             1.0.1
isort                 5.4.2
lazy-object-proxy     1.5.1
mccabe                0.6.1
mock                  4.0.2
more-itertools        8.4.0
packaging             20.4
pep8-naming           0.11.1
pip                   20.2.2
pluggy                0.13.1
py                    1.9.0
pycodestyle           2.6.0
pydocstyle            5.1.0
pyflakes              2.2.0
pylint                2.6.0
pyparsing             2.4.7
pyreadline            2.1
pytest                6.0.1
pytest-cov            2.10.1
pytest-repeat         0.8.0
pytest-rerunfailures  9.1
scspell3k             2.2
setuptools            49.6.0
```

colcon_cmake/task/cmake/__init__.py
```diff
@@ -206,6 +206,20 @@ def get_project_file(path, target):
...
+def get_msbuild_version():
+    """
+    Get the version of msbuild or throw an exception.
+
+    :rtype: str
+    """
+    cp = subprocess.run(
+        ['msbuild', '-version', '-noLogo'],
+        capture_output=True,
+        encoding=sys.getdefaultencoding(),
```
dirk-thomas, line 218: The `encoding` argument was only added in Python 3.6 but this package targets platforms still using Python 3.5 (Ubuntu Xenial, Debian Stretch).

colcon_cmake/task/cmake/build.py
```diff
@@ -148,25 +148,22 @@ async def _reconfigure(self, args, env):
         cmake_args += ['-DCMAKE_INSTALL_PREFIX=' + args.install_base]
         generator = get_generator(args.build_base, args.cmake_args)
         if os.name == 'nt' and generator is None:
-            vsv = get_visual_studio_version()
-            if vsv is None:
+            try:
+                msbuild_version = get_msbuild_version()
```
dirk-thomas, line 152: The current patch re-determines the `msbuild` version for each package which seem like unnecessary overhead. Instead it should determine it only once on the first invocation and afterwards reuse the cached result.

dirk-thomas (commented):
> the stuff in `test.py` and `__init__.py` was changed because the linter complained about it. I can revert those changes ...

Please do so to keep this patch minimal with only changes relevant to the topic of the PR..

dirk-thomas:
Please add a description why this change is being proposed in the first place.

rotu:
> Please add a description why this change is being proposed in the first place.

done

colcon_cmake/task/cmake/build.py
rotu: seems like premature optimization, but done.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -206,6 +207,20 @@ def get_project_file(path, target):
...
 
 
+@functools.lru_cache(maxsize=1)
+def get_msbuild_version():
+    """
+    Get the version of msbuild or throw an exception.
+
+    :rtype: str
+    """
+    cp = subprocess.run(
```
dirk-thomas, line 217: Nitpick: please don't abbreviate the variable name since it might not be obvious to everyone what it stands for.

colcon_cmake/task/cmake/build.py
```diff
@@ -148,25 +148,22 @@ async def _reconfigure(self, args, env):
...
-                    'please run within a Visual Studio Command Prompt.')
-            supported_vsv = {
-                '16.0': 'Visual Studio 16 2019',
-                '15.0': 'Visual Studio 15 2017',
-                '14.0': 'Visual Studio 14 2015',
-            }
-            if vsv not in supported_vsv:
+                    'msbuild is not found, '
+                    'please run within a Visual Studio Command Prompt.') from e
+            major_version = msbuild_version.split('.')[0]
```
dirk-thomas, line 157: Maybe the result should be converted to `int` for the following checks?

colcon_cmake/task/cmake/build.py
```diff
@@ -148,25 +148,22 @@ async def _reconfigure(self, args, env):
...
+            if major_version not in ('14', '15', '16'):
```
dirk-thomas, line 158: With the mapping `supported_vsv` removed I don't see a reason to whitelist explicit versions here. Just checking the lower bound (`14`) should be enough.

rotu: Totally agree. I'm not sure why the upper bound was in there in the first place, but I was keeping existing behavior.

dirk-thomas: > I'm not sure why the upper bound was in there in the first place, but I was keeping existing behavior.

Because the script passed the value of the `supported_vsv` dict to CMake and that value couldn't be computed by the input. You partially avoid the issue by only passing a substring of `-G` and hope that it is unique.

colcon_cmake/task/cmake/build.py
```diff
@@ -148,25 +148,22 @@ async def _reconfigure(self, args, env):
...
                 raise RuntimeError(
-                    'VisualStudioVersion is not set, '
-                    'please run within a Visual Studio Command Prompt.')
-            supported_vsv = {
-                '16.0': 'Visual Studio 16 2019',
-                '15.0': 'Visual Studio 15 2017',
-                '14.0': 'Visual Studio 14 2015',
-            }
-            if vsv not in supported_vsv:
+                    'msbuild is not found, '
```
dirk-thomas, line 155: This error message might be misleading in case the executable exists but fails for whatever reason. It would be good to distinguish these error cases to provide the correct information to the user.

colcon_cmake/task/cmake/build.py
dirk-thomas: I wouldn't call invoking a process potentially hundreds of times (for larger workspaces) for no reason "premature optimization"...

colcon_cmake/task/cmake/build.py
rotu: That error is only thrown on `FileNotFoundError`. Added some more explicit handling so the error message should always contain the entire version string.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -206,6 +207,20 @@ def get_project_file(path, target):
...
 
 
+@functools.lru_cache(maxsize=1)
+def get_msbuild_version():
+    """
+    Get the version of msbuild or throw an exception.
+
+    :rtype: str
+    """
+    completed_process = subprocess.run(
```
dirk-thomas, line 217: Please invoke `check_output` instead of `run` since that is what it is used for here.

colcon_cmake/task/cmake/__init__.py
```diff
@@ -206,6 +207,20 @@ def get_project_file(path, target):
     return project_file
 
 
+@functools.lru_cache(maxsize=1)
+def get_msbuild_version():
+    """
+    Get the version of msbuild or throw an exception.
```
dirk-thomas, line 213: The docblock should contain more information what kind of exceptions can be raised. This should use the `:raises ExceptionType: ...` style.

colcon_cmake/task/cmake/build.py
```diff
@@ -148,25 +148,27 @@ async def _reconfigure(self, args, env):
...
```
dirk-thomas, line 155: Nitpick: instead if "is"  it should be "was".

dirk-thomas (commented):
The function `get_visual_studio_version()` should be deprecated since it is no more used anywhere.

(I would have applied some of the changes I requested directly but it seems that editing the upstream branch is disabled.)

dirk-thomas:
@rotu Friendly ping.

dirk-thomas:
@rotu Friendly ping.

dirk-thomas:
Closing due to no response.

PR #98, opened 2020-12-06
Use command environment for determining MAKE args

cottsay:
Instead of grabbing the colcon process environment variables to determine what MAKEFLAGS will be when the process is invoked, grab the actual environment variables under which MAKE will be invoked.

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -287,7 +287,7 @@ class CmakeBuildTask(TaskExtensionPoint):
 
         The arguments are chosen based on the `cpu_count`, e.g. -j4 -l4.
 
-        :param env: a dictionary with environment variables
+        :param dict env: a dictionary with environment variables
         :returns: list of make arguments
         :rtype: list of strings
         """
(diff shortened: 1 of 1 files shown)


colcon_cmake/task/cmake/build.py
```diff
@@ -281,17 +281,18 @@ def _get_msbuild_environment(self, args, env):
...
         return env
 
-    def _get_make_arguments(self):
+    def _get_make_arguments(self, env):
         """
         Get the make arguments to limit the number of simultaneously run jobs.
 
         The arguments are chosen based on the `cpu_count`, e.g. -j4 -l4.
 
+        :param env: a dictionary with environment variables
```
dirk-thomas, line 290: Please add the parameter type here: `:param dict env: ...`.

dirk-thomas (changes requested):
Just for future readers: can you please briefly describe for which use case you needed this change.

colcon_cmake/task/cmake/build.py
cottsay: Nice catch: d3c9c53

cottsay:
> can you please briefly describe for which use case you needed this change

I'm experimenting with an extension which will augment the behavior of `make` by changing the `MAKEFLAGS`. I'm not sure yet whether this change is actually required for that effort, but is an improvement none the less, because it performs the check based on what variables will actually be passed to make, and not what the variables were when colcon was invoked.

dirk-thomas (approved):

PR #106, opened 2021-07-15
Add support for `Ninja Multi-Config`

KazNX:
The `Ninja Multi-Config` generator was introduced as part of CMake 3.17 supporting debug and release artefacts under the same `build` directory, like `Xcode` or `Visual Studio`. This change allows colcon to recognise `Ninja Multi-Config` as a mult-config generator and as a Ninja generator.

Note that in practice this change does not support an `apt-get` install of CMake on versions of current Ubuntu LTS releases - Unbutu 20.04 (latest current LTS) provided CMake version 3.16.

Change summary:

- Add `Ninja Multi-Config` to the set of generators where `is_multi_configuration_generator()` is `True`

### DIFF
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -55,7 +55,7 @@ async def has_target(path, target):
     generator = get_generator(path)
     if 'Unix Makefiles' in generator:
         return target in await get_makefile_targets(path)
-    if 'Ninja' in generator or 'Ninja Multi-Config' in generator:
+    if 'Ninja' in generator:
         return target in get_ninja_targets(path)
     if 'Visual Studio' in generator:
         assert target == 'install'
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -109,7 +109,7 @@ def get_buildfile(cmake_cache):
     """
     generator = get_variable_from_cmake_cache(
         str(cmake_cache.parent), 'CMAKE_GENERATOR')
-    if generator == 'Ninja':
+    if 'Ninja' in generator:
         return cmake_cache.parent / 'build.ninja'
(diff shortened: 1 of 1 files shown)


KazNX:
Checking failures. Not sure how to enforce these locally. I am running a linter.

KazNX:
I don't think the `flake8` failures are due to code I've introduced in the PR. Please advise.

KazNX:
PR has been rebased onto master fixing the CI check failures.

hidmic (approved):
> Note that in practice this change does not support an apt-get install of CMake on versions of current Ubuntu LTS releases - Unbutu 20.04 (latest current LTS) provided CMake version 3.16.

Hmm, IIUC the change should be transparent for earlier versions -- the generator simply won't ever match.

hidmic:
@dirk-thomas thoughts before merging?

KazNX:
> > Note that in practice this change does not support an apt-get install of CMake on versions of current Ubuntu LTS releases - Unbutu 20.04 (latest current LTS) provided CMake version 3.16.
> 
> Hmm, IIUC the change should be transparent for earlier versions -- the generator simply won't ever match.

This was something I was thinking about. The explicitly check for `Ninja Multi-Config` is fine because we only want to report this explicit generator as being a multi-condig ninja generator.

The compatibility issue is more to make sure `Ninja Multi-Config` is recognised as also being `Ninja`.

When I was looking, I found [this line](https://github.com/colcon/colcon-cmake/blob/6875f1dcb9109617cbd264fc60f6a2e488047f4c/colcon_cmake/task/cmake/__init__.py#L58).

Note the check is `if 'Ninja' in generator and not a straight equality test, so this will match both.

That said, in reviewing your comment I found [this line](https://github.com/colcon/colcon-cmake/blob/6875f1dcb9109617cbd264fc60f6a2e488047f4c/colcon_cmake/task/cmake/__init__.py#L112) is an equality test. Will fix.

PR #107, opened 2021-07-26
Add support for controlling the number of jobs used by some cmake tasks

KazNX:
This change adds support for `build` arg `--cmake-jobs` to set the number of jobs to use to preserve CPU for the OS. It also renames `_get_make_arguments()` to `_get_make_jobs_arguments()` to better reflect the responsibilities of the function.

There is currently support to set the number of job threads to the maximum available, _unless_  the `MAKEFLAGS` environment variable is set. The assumption is that `MAKEFLAGS` will be used by the build framework to control the number of jobs. However, this is only true for `make` - as far as I can tell, `ninja` does not have any such environment variable based control (related issues reported against ninja [922](https://github.com/ninja-build/ninja/issues/922), [1482](https://github.com/ninja-build/ninja/issues/1482)).

When `MAKEFLAGS` is not set, colcon will maxout the `-j` argument for ninja. Since ninja is so efficient, this can make the computer unresponsive for the OS and other applications. This change has been made to allow the OS to remain somewhat responsive.

As part of the change, I noted that `_get_make_arguments()` appears to be misnamed as the only thing it does is set the number of jobs to use, as evidenced by the use of the returned values. I've renamed it `_get_make_jobs_arguments()` to better reflect this. Happy to revert this change if required.

The `--cmake-jobs` args supports the following usage patterns:

- Positive number - use that many threads.
- Zero - use all available threads (current behaviour)
- Negative - use all but the absolute value of this number of threads (minimum use 1 thread).


KazNX:
PR has been rebased onto master fixing the CI check failures.

colcon_cmake/task/cmake/build.py
```diff
@@ -291,28 +300,37 @@ def _get_make_arguments(self, env):
...
+            )
+            matches = re.findall(regex, makeflags) or []
+            matches = [m[0] or m[1] for m in matches]
+            if matches:
+                # do not extend make arguments, let MAKEFLAGS set things
+                return []
+        # Use command line specified jobs if positive.
+        jobs_args = args.cmake_jobs if args.cmake_jobs is not None else 0
+        jobs = 0
+        if jobs_args <= 0:
```
hidmic, line 323: @KazNX hmm, isn't `args.cmake_jobs` ignored if it is positive?

hidmic (commented):
Overall this LGTM. @dirk-thomas ?

colcon_cmake/task/cmake/build.py
```diff
@@ -281,7 +290,7 @@ def _get_msbuild_environment(self, args, env):
             env['CL'] = ' '.join(cl_split)
         return env
 
-    def _get_make_arguments(self, env):
+    def _get_make_jobs_arguments(self, args, env):
```
hidmic, line 293: @KazNX hmm, this is no longer just `make` arguments.

colcon_cmake/task/cmake/build.py
```diff
@@ -291,28 +300,41 @@ def _get_make_arguments(self, env):
...
+            with suppress(AttributeError):
+                # consider restricted set of CPUs if applicable
+                jobs = min(jobs, len(os.sched_getaffinity(0)))
+            if jobs is None:
+                # the number of cores can't be determined
+                return []
+            # Finalize jobs as as CPU count deducting the limit specified.
+            jobs = max(jobs + jobs_arg, 1)
+        else:
+            jobs = jobs_arg
```
hidmic, line 337: @KazNX nit^2: minor simplification 
```suggestion
        jobs = 0
        if args.cmake_jobs is not None:
            jobs = args.cmake_jobs
        # If positive, use jobs as is, even if it's more than available.
        # Excessive jobs specified is a user error.
        if jobs <= 0:
            # Base off the number of CPU cores if jobs arg non-positive.
            cores = os.cpu_count()
            with suppress(AttributeError):
                # consider restricted set of CPUs if applicable
                cores = min(cores, len(os.sched_getaffinity(0)))
            if cores is None:
                # the number of cores can't be determined
                return []
            # Finalize jobs as as CPU count deducting the limit specified.
            jobs = max(cores + jobs, 1)
```

colcon_cmake/task/cmake/__init__.py
```diff
@@ -139,6 +139,27 @@ def get_generator(path, cmake_args=None):
     return generator
 
 
+def is_jobs_base_generator(path, cmake_args=None):
```
hidmic, line 142: @KazNX nit: would `is_jobs_based_generator` be more accurate? Same elsewhere.

colcon_cmake/task/cmake/build.py
```diff
@@ -281,38 +290,53 @@ def _get_msbuild_environment(self, args, env):
...
 
-    def _get_make_arguments(self, env):
+    def _get_jobs_arguments(self, args, env):
         """
         Get the make arguments to limit the number of simultaneously run jobs.
 
         The arguments are chosen based on the `cpu_count`, e.g. -j4 -l4.
 
+        This only handles some generators, specifically; Makefiles and Ninja
+        derivatives.
```
hidmic, line 300: @KazNX nit: let's keep it generic
```suggestion
        This only handles jobs based generators, see `is_jobs_based_generator()`.
```

hidmic (approved):
This looks great to me, but I'd appreciate @dirk-thomas or @cottsay input as well.

cottsay (commented):
Doesn't simply specifying the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable work for Ninja?

colcon_cmake/task/cmake/build.py
```diff
@@ -63,6 +64,12 @@ def add_arguments(self, *, parser):  # noqa: D102
             '--cmake-force-configure',
             action='store_true',
             help='Force CMake configure step')
+        parser.add_argument(
+            '--cmake-jobs',
+            type=int,
+            help='Number of jobs to use for supported generators (e.g., Ninja '
+            'Makefiles). Negative values subtract from the maximum '
+            'available, so --jobs=-1 uses all bar 1 available threads.')
```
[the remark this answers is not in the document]
KazNX: Not a typo

KazNX (commented):
> Doesn't simply specifying the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable work for Ninja?

@cottsay Not sure, but it was introduced in CMake 3.12. I'm working with baseline support for 3.10 as ships with Ubuntu 18.04.

KazNX:
@cottsay Looking at the [CMake 3.12 release notes](https://cmake.org/cmake/help/latest/release/3.12.html#command-line), it looks like they added `--parallel` and `--jobs` arguments as well as `CMAKE_BUILD_PARALLEL_LEVEL`. This is great, but I need to target a minimum version of CMake 3.10 for now (the default CMake version for Ubuntu 18.04).

This changes the context of this PR a bit and I see the need to revise the original code I modified to deal with that as well. I propose the ideal solution is to use the changes I've made with CMake < 3.12. For CMake 3.12+, change this function to return `--parallel` and `--jobs` instead, which special provisions not to append these after `--` is added to the command line. `_is_jobs_based_generator()` can then return `True` in call cases.

My question is, are the changes better made in this PR or a to push this through and start a new PR?

cottsay:
> are the changes better made in this PR or a to push this through and start a new PR?

I see two conceptual changes:
1. We should to stop mucking with the `MAKEFLAGS` if `CMAKE_BUILD_PARALLEL_LEVEL` is specified in the environment, which would support this use case for any system with CMake 3.12 or newer. This should be a trivial change. We should *not* remove the existing `MAKEFLAGS` arguments outright in favor of `CMAKE_BUILD_PARALLEL_LEVEL` because it uses `--load-average`, which achieves different behavior than `--jobs` alone.
2. We should introduce a mechanism to support your use case with CMake older than 3.12. I'm not sure if a full-fledged command line option is the right move or something a little more "hidden" like an environment variable. Maybe a deprecated command line option would be good - or one that shows a deprecation warning if you're already using 3.12 and should switch to the supported environment variable for doing this.

The latter can be done in this PR, I think. The former should probably be done in a new PR. Thoughts?

KazNX:
> I see two conceptual changes:

I fully agree with point 1, but would add that it's not just using `CMAKE_BUILD_PARALLEL_LEVEL`. It needs to be either/or with passing cmake args for `--parallel` and/or `--jobs`. I forsee some annoying permutations and need to learn more about how these interact.

On point 2, the deprecation warning only makes sense if `CMAKE_BUILD_PARALLEL_LEVEL` or `--parallel` and/or `--jobs` are present. Something like:

```
if cmake_ver >= 3.12 and (have_env_var('CMAKE_BUILD_PARALLEL_LEVEL') or have_cmake_arg('--parallel') or have_cmake_arg('--jobs')):
  print('deprecation warning')
```

Again, I need more information aboug how the command line arguments interact with `CMAKE_BUILD_PARALLEL_LEVEL`. Will get back to you on that.

KazNX:
FYI, `-j` `--parallel` and `CMAKE_BUILD_PARALLEL_LEVEL` are all the same. Command line help from CMake 3.12;

>  -j [<jobs>] --parallel [<jobs>] = Build in parallel using
                   the given number of jobs. If <jobs> is omitted
                   the native build tool's default number is used.
                   The CMAKE_BUILD_PARALLEL_LEVEL environment variable
                   specifies a default parallel level when this option
                   is not given.

KazNX:
@cottsay @hidmic 
I've made a significant revision of this PR based on CMake 3.12+ supporting `--parallel` or `-jN`. Note that it sill uses a new `colcon build` argument `--cmake-jobs` in all cases. I can't use `--cmake-args` because that only affects CMake configuration and we need to inject arguments into `cmake --build`.

The changes are summaried as:

- For cmake 3.12+, we use `cmake --build -jN`.
- For cmake < 3.12, we use `cmake --build [args] -- -jN -lN`, but only for Ninja and Makefiles generators.

Note we no longer check if we have a jobs based generator. This is now managed internally to `_get_jobs_arguments()`.

Also note we now add `--` inside `_get_jobs_arguments()` as it is not needed for CMake 3.12+

dirk-thomas:
Since I don't have the spare time to review this change I leave it up to the co-maintainers to review and test this.

A general - fairly significant - concern I have: the Travis CI integration for all `colcon` repositories has been broken for a while. Merging any changes has an increased risk to introduce regressions. I strongly suggest to first "get the house in order" and reestablish CI coverage for the lost platforms - e.g. by adding GitHub action support to all repos.

cottsay:
>  ...over the current state of master rather than a applying to this PR.

Neither this PR nor master are getting test coverage on Linux or macOS because of the missing Travis builds.

> are there any further changes required to get this PR merged?

I'm still not feeling great about merging a PR that adds a command line option to support such a narrow use case, especially one that has an official mechanism available in newer versions of CMake.

Here's a completely different suggestion - could we expose a free-form `--cmake-build-args` option to correspond to the existing free-form `--cmake-args` option? That would be useful outside of the Ninja use case, and wouldn't be obsoleted by CMake 3.12+.

KazNX:
> Here's a completely different suggestion - could we expose a free-form --cmake-build-args option to correspond to the existing free-form --cmake-args option? That would be useful outside of the Ninja use case, and wouldn't be obsoleted by CMake 3.12+.

Yes, I like this option. It does mean removing the `_get_jobs_arguments()` method (previously `_get_make_arguments()`) completely because to prevent it from adding *implied* `-j` arguments.

KazNX:
I've created a new PR for the `--cmake-build-args` approach. See here https://github.com/colcon/colcon-cmake/pull/110

I managed to maintain `_get_make_arguments()`, though I think that functionality is questionable.

We can deprecate this PR if the 110 is the preferred solution.

PR #133, opened 2023-09-06
Use packaging.version in place of pkg_resources.parse_version

cottsay:
This is an API breaking change, but I don't really see a good way to avoid it.

@KazNX FYI

### DIFF
--- a/colcon_cmake/task/cmake/__init__.py
+++ b/colcon_cmake/task/cmake/__init__.py
@@ -9,7 +9,7 @@ import sys
 
 from colcon_core.environment_variable import EnvironmentVariable
 from colcon_core.subprocess import check_output
-from pkg_resources import parse_version
+from packaging.version import Version
 
 """Environment variable to override the CMake executable"""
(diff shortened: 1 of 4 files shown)


nuclearsandwich (approved):
The API break mentioned here is due to the return type changing in response right? How much duck typing overlap is there between the old and new type? This is a break and so worthy of a major bump but how much impact do we anticipate downstream?

cottsay:
> The API break mentioned here is due to the return type changing in response right?

Affirm.

> How much duck typing overlap is there between the old and new type?

Probably a lot. I think maybe `pkg_resources` was just using a bundled version of `packaging`.

> This is a break and so worthy of a major bump but how much impact do we anticipate downstream?

I anticipate no breakage. This function was introduced in #57 to support a specific feature and isn't used in any supported colcon packages outside of `colcon-cmake` itself. We're pre-1.0 here, and `colcon-core` should lead the 1.0 drive, so I think a minor bump will be sufficient.

PR #145, opened 2024-11-21
Switch implicit make flags to use GNUMAKEFLAGS var

cottsay:
There are a few notable elements addressed by this change:
1. The --load/-l argument is GNU-only and doesn't work on BSD make
2. The GNUMAKEFLAGS environment variable was previously ignored when looking for existing make arguments
3. By passing the flags via command line arguments to CMake's build invocation, the makeflags were sometimes omitted due to incompatibility.

Until I can come up with some tests for this, I'll keep this PR in draft.

### DIFF
--- a/colcon_cmake/task/cmake/build.py
+++ b/colcon_cmake/task/cmake/build.py
@@ -84,6 +84,8 @@ class CmakeBuildTask(TaskExtensionPoint):
         if environment_callback is not None:
             environment_callback(env)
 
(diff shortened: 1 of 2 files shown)


nuclearsandwich (commented):
Initial review looks reasonable. The inversion of get -> set means that we lose an internal API surface to check these values for downstream processes but I have a feeling that we'll come up with better internal APIs when we have a use case to drive them.

PR #156, opened 2026-01-03
Fix packaging deprecation and flake8 regression in tests

MatthijsBurgh:


MatthijsBurgh:
@cottsay Happy new year. Please review this PR.

christophebedard (approved):
I don't think I have permissions on this repo, but here's a ✔️ anyway

MatthijsBurgh:
@christophebedard @cottsay please approve the workflow

test/test_parse_cmake_version.py
```diff
@@ -33,7 +33,7 @@ def test_parse_cmake_version():
             # Input string was garbage. Assert parsing failed.
             assert parsed_version is None
         else:
-            assert parsed_version._version.release[0:3] == expected_version
+            assert parsed_version.release[0:3] == expected_version
```
cottsay, line 41: This seems right, thanks :+1:

That property [was added](https://github.com/pypa/packaging/commit/0cd72c3b20f7b716caba0d221f74405b2531c7d4) in version 17.0 of `packaging`, released in February of 2018. While the test will fail if the property doesn't exist, the underlying code still functions with older versions of `packaging`. Of note, RHEL 8 (via EPEL) has 16.8 (for ROS 2 Humble), and Ubuntu Xenial was 16.6 (for ROS 2 Ardent, which is EOL).

I considered whether to bump the minimum version in `setup.cfg` and `stdeb.cfg`, but I think skipping the test on those relatively ancient versions is probably better. I'll push a quick change to do so.

Back-linking to related PRs #57 and #133.

cottsay (approved):

